PR #152, opened 2015-10-25
Fix arguments propagation to html_of_md

venator:
This patch adds optional arguments propagation to recursive calls of `Omd_backend.html_of_md`. It makes `html_of_md` behaviour more consistent, in particular when using the `?override` argument, since only markdown top level elements were affected by `override`.

The following arguments are propagated:

```
~override ~pindent ~nl2br ~cs
```


pw374:
Thanks!

PR #163, opened 2017-08-10
Update opam & be compatible with safe-string

Chris00:


pw374:
thank you! Not sure why I didn't see this one earlier...

PR #167, opened 2018-08-21
Cleanup codebase, remove code duplication, switch to dune

nojb:
This PR contains some patches I wrote almost a year back but somehow got lost in the shuffle. In short, it:

- Removes much of code duplication of the codebase (see commits starting at "Start simplification of `Omd_representation.token`")
- Removes generated code by simple use of functions
- Switches on `-safe-string`
- Removes a bunch of warnings
- Removes dead code
- Switches to `dune` for building

All this cleanup amounts to almost 4K lines of code removed.

Because I wrote this a long time back I forgot some of the details, so this PR could benefit from some kind of review. However, IMHO, the current codebase is in desperate need of love, so even if there is a risk of some regression, it may end up being a good idea to merge this PR.

NB: since there was such a great deal of code refactoring to do, many intermediate commits do not compile (eg those titled "Continue").

About the tests:

- They can be launched with `dune runtest`, but they still use a "hand-written" test harness; the next step is to switch them to "`dune`" tests so that we can use `dune promote`, etc.
- Many of them fail (as far as I remember this is the case in `master` currently), but a subset of them (maybe even all?) are in fact correct so it is just a matter of going through them and committing an updated result.


omd.opam
```diff
@@ -0,0 +1,22 @@
...
+authors: [ "Philippe Wang <philippe.wang@gmail.com>" ]
+license: "ISC"
+homepage: "https://github.com/ocaml/omd"
+dev-repo: "https://github.com/ocaml/omd.git"
+bug-reports: "https://github.com/ocaml/omd/issues"
+tags: [ "org:ocamllabs" "org:mirage"  ]
+depends: [
+  "base-bytes"
+  "base-bigarray"
+  "dune" {build & >= "1.0"}
```
mseri: L’attributo lang 1.1 usato in dune-project è compatibile con dune 1.0?

nojb:
Good catch! I'm not sure, but I guess not. The `dune-project` file was
auto-generated and I missed it. Will fix.

omd.opam
nojb: Thanks, fixed.

src/omd_lexer.mli
```diff
@@ -12,13 +12,13 @@ val lex : string -> t
     in highest priority whereas functions in [extensions] are applied
     with lowest priority. *)
 
-type bigstring = (char,
-                  Bigarray.int8_unsigned_elt,
-                  Bigarray.c_layout) Bigarray.Array1.t
+type bigstring = (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Array1.t
```
Chris00: Please have lines of at most 80 chars:
```
type bigstring =
  (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Array1.t
```

Chris00:
I am glad to see some work on this library.  A good test is to try to build the website with the updated version...

nojb:
I am not familiar with the procedure to do that. Could you, or someone else, give it a shot and report back the results? Thanks!

nojb:
> I am glad to see some work on this library. A good test is to try to build the website with the updated version...

I can give this a try if someone can point to instructions somewhere to do it.

nojb:
I managed to generate the site. Fixed two bugs in the process. See [`ocaml.org`](https://nojb.github.io/omd/ocaml.org/).

I haven't scrutinized the output much. Let me know if you spot any issues! Thanks!

Chris00:
It seems double quotes are transformed into single quotes provoking many errors when executing in the toplevel. See e.g. https://nojb.github.io/omd/ocaml.org/learn/tutorials/modules.html

nojb:
I pushed an extra commit switching some tests to "dune tests". I have more stuff in the pipeline, but I would rather wait until this PR is merged before moving forward with those.

Chris00:
@pw374 Could you please review the code?

nojb:
@pw374 Thanks for the review.

Regarding the style/indentation changes, I used `ocp-indent` to reindent the files. The style I used is a fairly standard one, but I understand that it may not be to your liking. Personally I find it more readable.

- For reviewing purposes, you can direct github not to show whitespace changes; did you try that?
- If wanted, I can reindent with a style closer to the one used in the existing codebase.
- I agree that the changes in spacing around colons are a bit capricious. As you mention working on this codebase is quite hard and some changes were made "in the flow of things".

Apart from stylistic changes, the only other point you raise is the possible impact on performance due to the change in the representation of tokens. I guess you are referring to a potential increase in allocations, but unless we are thinking of very resource-constrained execution environments (are you?), I don't think it would be a big deal for the OCaml GC.

src/omd_representation.mli
```diff
@@ -51,100 +51,69 @@ and href = string
...
-    Notably, this allows to use the property that in the match
-    case [Ampersand _ ->], we know there are at least 2 ampersands.
-    This is particularly useful for some characters, such as newlines
-    and spaces. It's not useful for all of them indeed but it has
-    been designed this way for the sake of uniformity (one doesn't
-    want to know by heart which constructor have that "at least 2"
-    property and which haven't).
-*)
+      The integers carried by constructors means that the represented
+      character appears (n+2) times. So, [Ampersand(0)] is "&&".
```
pw374: I think this comment has to be updated (or deleted), as it becomes badly out of date following the conversion to `Delim of int * delim`, right??

pw374:
Let's merge asap... like now, otherwise it could be delayed for months for bad reasons. Worst case scenario, we revert the changes...

Now my biggest concern here is that I don't wanna look at each refactoring part about `XXX | XXXs of int` being converted to `Delim of int * delim`, because there are so so many of them and it can easily be very tricky. 

P.S. I'm interested in knowing how to hide whitespace changes in Github. I tried the option "Hide whitespace changes; Apply and reload" on the "Files changed" tab, but it's not useful at all...

PR #169, opened 2018-10-12
Use wrapped library

nojb:
This PR switches the build of the library to the "modern" style, using module aliases. This breaks the public API of the library (`Omd_X` becomes `Omd.X`), but it has a number of advantages going forward.

@pw374


pw374:
OK... Thanks.

I'm not a fan of this change, because it breaks compatibility everywhere... but OK.

PR #171, opened 2018-10-12
Add CommonMark test suite

nojb:
This PR adds adds the CommonMark test suite (https://commonmark.org/). Most of the tests fail currently, but my intention is that my ongoing refactoring will allow me to improve the state of affairs.

There is also a script to re-import the test suite from upstream.

@pw374


pw374:
thanks!

PR #172, opened 2018-10-12
Add Travis script

nojb:
This PR adds a simple Travis script.

Since most of the tests are currently failing the script only tests that the library compiles. Tests will be enabled later.

@pw374


pw374:
awesome! thanks

PR #174, opened 2018-11-12
Refactor library, add CommonMark compatibility

### LINKED ISSUE #40: Please document omd_representation.mli
In particular, the meaning and the (assumed) constraints on the integers should be described.

### LINKED ISSUE #51: Omd.t: express more invariants
I think it would be nice to change the representation of `Omd.t` (possibly using polymorphic variants) to express that certain tags (e.g. block elements) cannot be inside other ones.

I agree this is a big change but IMHO it may be worth.  It can also emphasize the possibilities of this library.  For example, block level elements inside blockquotes seem to be correctly handled which is not the case of pandoc.

### LINKED ISSUE #76: Remove `Ulp` and `Olp`?
I was wondering whether `Ulp` and `Olp` are still needed.  One could indeed move `make_paragraphs` into `omd_parser` and make the transformation right on the spot instead of using the label `Ulp` (it seems that, once created, one does not modify the content `Ulp` except to make paragraphs).

Maybe it is just me but IMHO `Ulp` and `Olp` add a conceptual burden to somebody wanting to use the library (and often make for redundant cases when processing the AST).

### LINKED ISSUE #94: Parsing of list + code + header
This is related to https://github.com/pw374/omd/issues/92 but on the parsing side. IMHO,

<pre>
- A
- B

    ```
    code
    ```
# header
</pre>

should parse as 

<pre>
[Ulp [[Paragraph [Omd.Text "A"]];
         [Paragraph [Omd.Text "B"]; Code_block ("", "code")]];
 H1 [Text "header"]];
</pre>

(as Pandoc but unlike Github).

### LINKED ISSUE #121: indentation in Markdown
https://github.com/ocaml/omd/issues/60#issuecomment-51192924 made me realise that the 4-space rule in Markdown should perhaps be "fixed"...

Pandoc and Github loosely and badly implement the 4-space rule for list indentation and it's awful, see for instance how https://github.com/pw374/sandbox/raw/master/lists.md renders on github https://github.com/pw374/sandbox/blob/master/lists.md.

OMD gives a rather sane semantics for lists to prevent those insane behaviours.

Now I'm wondering if the same (or a similar) treatment should be applied to other elements...

Any thoughts?

### LINKED ISSUE #123: Html v.s. Html_block
Before, when HTML was treated opaquely, the difference between `Html` and `Html_block` made sense.  Now that HTML is parsed, I think it is the job of `Omd.to_markdown` to distinguish between the two (the user should not need to care).  Of course, exposing a function `Omd.is_block_tag` (or similar) is welcome.

### LINKED ISSUE #126: support CommonMark
stmd seems to be

> a spec for "standard markdown," with matching C and javascript implementations

https://github.com/jgm/stmd

It might make sense (or not) to support it at one point

### LINKED ISSUE #127: Verify new list of accepted HTML tags
`html`, `head`, `body`, `title`, `style`, `link` and `meta` have been accepted since https://github.com/ocaml/omd/commit/78e4708b9f1553dbeca50ece12516ed6c007bf2e.

That makes sense thanks to `media:type="text/omd"` and because the HTML parser was rewritten to be a lot more powerful than the one used when that feature was introduced.

Now someone (perhaps me, but not right now) should check that the few lists of HTML tags are what they should be.

### LINKED ISSUE #129: rewrite backends
Once https://github.com/ocaml/omd/issues/128 has been dealt with, the backends should be rewritten.

### LINKED ISSUE #132: Using 'opam' in the Makefile makes it hard to package with other systems
The Makefile seems to assume the presence of 'opam', which won't be true everywhere.

### LINKED ISSUE #137: Improper escaping in Markdown output
```
# let md =
let md =
    [Omd.Html_block ("pre", [],
       [Omd.Html ("code", [("class", Some "ocaml")],
          [Omd.Html ("span", [("class", Some "comment")],
             [Omd.Text "("; Omd.Text "*"; Omd.Text " "; Omd.Text "Hello";
              Omd.Text " "; Omd.Text "*"; Omd.Text ")"])])])];;
          val md : Omd.element list =
  [Omd.Html_block ("pre", [],
    [Omd.Html ("code", [("class", Some "ocaml")],
      [Omd.Html ("span", [("class", Some "comment")],
        [Omd.Text "("; Omd.Text "*"; Omd.Text " "; Omd.Text "Hello";
         Omd.Text " "; Omd.Text "*"; Omd.Text ")"])])])]
# Omd.to_markdown md;;
- : string =
"<pre><code class='ocaml'><span class='comment'>\\(\\* Hello \\*\\)</span></code></pre>"
```

### LINKED ISSUE #154: What is the difference between Ul and Ulp?
The abstract representation of Markdown in `omd.mli` provides `Ul` and `Ulp` for unnumbered lists but both are isomorphic. What is the purpose of having two?

```
type element =
| H1 of t          (** Header of level 1 *)
...
| Ul of t list     (** Unumbered list *)
| Ol of t list     (** Ordered (i.e. numbered) list *)
| Ulp of t list
| Olp of t list
...
```

nojb:
(**not yet for merge**, see below)

This PR is the second part of #167. I had two objectives when I started working on this library:

- Make the library hackeable
- be [CommonMark](https://spec.commonmark.org/)-compartible

The current PR mostly attains these two objectives. Unfortunately, along the way, virtually the whole library had to be rewritten, so almost none of the original code is left. This was not my initial intention, but it was the result of incremental changes.

There are two main changes in this PR:

- A new representation for markdown documents:
```ocaml
type 'a link_def =
  {
    label: 'a;
    destination: string;
    title: string option;
  }

type 'a block =
  | Paragraph of 'a
  | List of list_kind * list_style * 'a block list list
  | Blockquote of 'a block list
  | Thematic_break
  | Heading of int * 'a
  | Code_block of (fenced_code_kind * string) option * string option
  | Html_block of string
  | Link_def of string link_def

type link_kind =
  | Img
  | Url

type inline =
  | Concat of inline list
  | Text of string
  | Emph of emph_kind * emph_style * inline
  | Code of string
  | Hard_break
  | Soft_break
  | Link of link_kind * inline link_def
  | Ref of link_kind * inline * string link_def
  | Html of string

type t = inline block
```
This new representation can only represent "legal" markdown documents, which removes a whole category of bugs.
- A new parsing algorithm, corresponding to the representation of markdown documents in two levels (blocks and inlines).
  - First, the block structure is parsed with the inline contents taken as strings, and second, the inline content is parsed within each leaf block (this is the most complicated part of the code).
  - The new parser does not lex the input data into tokens, it parses the strings directly at the level of characters.
  - It would be easy to modify the parser for full Unicode support (right now, the input is considered to be a stream of bytes and that is enough for 99% of the cases, but anyway).

After much iterative work, the library passes roughly 720 out of 735 tests. The remaining tests concern some tricky edge cases which I intend to fix before the merge or shortly after. Because of the iterative way I developed this library there is a fair amount of cleaning up and simplifications that I will be doing in the next few days. Once that is done, I think we should merge this PR.

I will also in the next few days try to regenerate `ocaml.org` to start looking for regressions there. Any help would be appreciated (cc @Chris00)

Fixes #40, fixes #51, fixes #76, fixes #94, fixes #121, fixes #123, fixes #126, fixes #127, fixes #129, fixes #132, fixes #137, fixes #154


nojb:
@pw374

nojb:
One important point: I removed the support for **extensions** from this version. I want to add it back, but it needs to be redesigned to work with the new representation of markdown documents; I would like to hear about different use-cases before doing that.

pw374:
Woooow :)

That's a huge PR!

I'm not sure my previous version tagging (2.0.0.alpha) is still relevant. We're gonna jump from 2.0.0.alpha to 3.0.0.alpha! haha

Option `-otoc` is used for ocaml.org and not supporting it will break the website pretty hard I think.
Option `-r` is used as well, for highlighting code snippets.

Do you support `media:type="text/omd"` or did you remove that as well?

nojb:
> Option -otoc is used for ocaml.org and not supporting it will break the website pretty hard I think.

I will add this back shortly.

> Option -r is used as well, for highlighting code snippets.

I don't remember what this is, but will add it back if necessary.

> Do you support media:type="text/omd" or did you remove that as well?

Could you remind me what is this about?

pw374:
> > Option -otoc is used for ocaml.org and not supporting it will break the website pretty hard I think.
> 
> I will add this back shortly.

OK 👍 

> > Option -r is used as well, for highlighting code snippets.
> 
> I don't remember what this is, but will add it back if necessary.

It allows to plug a code highlighter to sections of code, using the language information.
When you write
````
```ocaml
let r = 42
```
````
there's syntactic highlighting in Github. OMD doesn't know how to highlight anything but it can delegate to an external command, using option `-r`

> > Do you support media:type="text/omd" or did you remove that as well?
> 
> Could you remind me what is this about?

In markdown when you switch to HTML (in other words, when you enter an HTML section), you can either continue to parse it as Markdown or stop and consider it a block of HTML. The 2 choices are very different. I'm not sure what people do nowadays.

But in OMD it was concluded (back then) that when we enter an HTML block, there would be no Markdown parsing inside, except for HTML nodes that would have attribute `media:type="text/omd"`;  that was the best compromise we could find.

ostera:
Any updates on this? 🙈

nojb:
Not yet - but it is at the top of my list, as soon as I get some free time.

avsm:
In the interests of unblocking this, I propose merging it to trunk and constraining the ocaml.org build to use the opam-released version for now.  That'll make it significantly easier to try the trunk version against other tools such as `mdx` that currently call out to pandoc, while not affecting our infrastructure builds for the website at all.  Does that sound ok to you @pw374?

pw374:
That’s totally ok to me

avsm:
Thanks for the quick response @pw374!  I'll just add in some CI before merging this; Travis seems to have become disconnected.

nojb:
To repeat what I told @avsm, I am OK with it as well, but I would prefer to take one last look at the code before merging, just to refresh my memory in case there was something urgent left to do. I can do this over the weekend, and merge afterwards.

Sorry for stalling on this PR!

avsm:
Sure thing @nojb -- very much appreciate your hard work on the PR to date!

nojb:
OK, it took me a week, but I finally took a look at what's there. Roughly these are the outstanding issues:

- Parsing: the parsing of Markdown is two-stage: first one decodes the block structure (paragraphs, lists, list items, quotes, etc.) and then one has to parse the inline structure inside each block (emphasis and links). The first part of the parsing can process the content line-by-line and is fairly simple. The inline part is rather tricky. Both parts are now using some variation of combinator parsing that get the job done, but am not very happy with. Maybe a combination of regular expressions and code could work better. Definitely something that needs looking at. More importantly the logic that parses the inline structure is way too complicated and needs to be simplified.

- The Markdown->{Text,Markdown,HTML,...} converters need some more work. Only the HTML one is complete as far as I know (it is needed for the tests).

- UTF-8 support is missing.

- The driver program has been stripped of almost all its options and needs to be rebuilt using the new Markdown representation. It should be easy.

- A couple of tests (with tricky inline structure) fail. This is related to the first item above.

That's all I could see. I will merge the PR since I don't think we gain much from it by letting it languish, but further work is needed to get the whole package up to speed.

avsm:
I've opened up separate issues for the remaining regressions that @nojb pointed out above.

Issue #178, opened 2019-04-16
Tests are broken (wrong outputs and compilation errors)


clecat:
Hello,
I have been trying to do some changes to omd, but first I attempted to run the tests, and it appeared that they were broken:

In particular,the last commit modified the Code_block type from:
> (fenced_code_kind * string) option * string option

to this
> (fenced_code_kind * (string * string)) option * string option

But the test_spec.ml file was not updated with those modifications.
Before doing a PR for it, I was wondering if all the tests had to pass or if some specs are currently not implemented ?

Indeed the specs 391, 394, 406, 446, 447, 491 and 494 are not passing, in addition of the tests 'blockquote + header + par', 'blockquote + code (tab)' and the test of _link-with-code.html_.

Are those errors some regressions or were them ever passing ?

nojb:
Hello, thanks for your contribution!

- Indeed, `test_spec.ml` needs to be adapted to the last commit.

- Note that this test and the ones in `tests/cow` are old and do not test a lot of features, so I was thinking about getting rid of them and keeping only the official commonmark spec tests (in `tests/commonmark`).

- Indeed, there is a small handful of the official tests that do not pass (they have to do with Unicode support and some edge cases of emphasis parsing). They can be disabled so that we have a fully passing testsuite.

In summary, my suggestion is to disable `test_spec.ml` and the tests in `tests/cow` and the small handful of non-passing commonmark spec tests. We can decide later if we get rid of the old tests or we fix them to pass.

Let me know if you have any questions.

clecat:
Thank you for the answer, by getting rid off/disabling, do you deleting the old tests (test_spec.ml and tests/cow) and disabling the commonmark ones, or just disabling them all, keeping the old tests ?

nojb:
For now, we can just disable them (by commenting out the relevant `dune` fragments), and we can decide whether to remove or update them later.

nojb:
Fixed by #179

PR #179, opened 2019-04-16
Corrected compilation issues in tests, disabled broken tests

clecat:
This is a small pull request solving the problems mentioned (here)[https://github.com/ocaml/omd/issues/178]:
- The compilation errors have been solved, updating the code to the latest version.
- I commented the broken test linked to the absence of implementation
- I added a runtest_all rule (call it with ` dune build @runtest_all `) which launches every test (even the broken one mentioned before).


nojb:
LGTM. I will look to get the CI enabled again since that will help for future PRs.

nojb:
@avsm has re-enabled the CI for this repository, so future PRs will benefit from the changes you just made here. Thanks!

clecat:
No problem :) One thing to keep in mind though (I just saw it), the dune.inc files are generated, which means that the changes I did will not be kept after a refresh. I will try to solve it later so the changes are permanent (keeping somewhere a blacklist of unit tests or something like that), but it does not seems urgent, as apparently the files were only generated once since the beginning

nojb:
> No problem :) One thing to keep in mind though (I just saw it), the dune.inc files are generated, which means that the changes I did will not be kept after a refresh. I will try to solve it later so the changes are permanent (keeping somewhere a blacklist of unit tests or something like that), but it does not seems urgent, as apparently the files were only generated once since the beginning

Indeed, that file is only regenerated when importing the commonmark test suite, which does not happen very often. So yes, we can leave this for later.

PR #180, opened 2019-04-16
Corrected linguist language detection

clecat:
Just a little PR for the sake of clarity:

git-linguist detects the dune.inc files as php files, distorting the stats of the project, making it shown as mainly coded in php (~44,7% of php).

Kinda funny considering the "pure ocaml" mention :p

This pull-request does not change directly the language of .inc files, but as dune.inc files are auto-generated, it adds a git attribute, telling linguist to ignore those files


nojb:
LGTM

Issue #175, opened 2019-04-05
Add support for section wrapping of header identifiers


clecat:
Hello,
I'm currently working on [ocaml-mdx](https://github.com/realworldocaml/mdx) which uses pandoc in order to generate html code from the markdown files.
I want to replace it for omd in order to avoid a Sys.command call and for more portability.
But the problem is that we are using the "--section-divs" argument, which does the following:
> Wrap sections in \<section> tags, and attach identifiers to the enclosing \<section> rather than the header itself.

Generating this:
```
<section id="test0--toto--and-titi" class="level1">
<h1>Test0, -toto- and titi</h1>
<p>toto</p>
</section>
```
Instead of this:
```
<h1 id="Test0-toto-and-titi">Test0, -toto- and titi</h1>
<p>toto</p>
```

As far as I know, omd does not currently support such thing.
Does it seems reasonable to add support for it ? (in the binary and the library API)

nojb:
`omd` was recently rewritten (#174, no release yet) but the new API should allow to add this without much trouble. You would need to do 2 easy things:
1) Rewrite the implementation of the HTML printer to use an open recursion style (https://github.com/ocaml/omd/blob/master/src/html.ml#L135) so that individual cases can be overriden by the library client code, and
2) in `ocaml-mdx` you would need override the `heading` method of the HTML printer to keep track of the section scope and insert the `section` tags as needed.

Neither of the two items are particularly difficult and they would be a nice intro if you are interested to dig a little deeper into `omd`. I am glad to help you along, but I don't have time at the moment to work on them myself.

clecat:
Nice thanks, I will give it a try as soon as possible :)

clecat:
Okay so, does using a functor looks to you ?
Something like that:
We have a base module containing a dumping function for every case of the ast.
We use it for generating a new To_html module via a functor.
And a user who wants to override a function simply creates a module including the base module and redefining the function.

nojb:
That's one possibility. You could also use objects. My preferred approach, however, is to be low-tech and just use records:
```ocaml
type printer =
  {
    block: printer -> Buffer.t -> inline block -> unit;
    paragraph: printer -> Buffer.t -> inline -> unit;
    blockquote: printer -> Buffer.t -> inline block list -> unit;
    (* rest of cases for blocks *)
    inline: printer -> Buffer.t -> inline -> unit;
    html: printer -> Buffer.t -> string -> unit;
    (* rest of cases for inlines *)
  }
```

[...]

PR #185, opened 2019-04-30
Added printer implementation for dumping modifications from the user

clecat:
This pull request is part of the work needed for this [issue](https://github.com/ocaml/omd/issues/175):
In order to modify the behaviour of omd from the user, I added a printer record, containing the printing functions.
It allows the user to redefine their behaviour by creating a new record with the new functions, then give it as an argument for the to_html function.

For example, the following code wraps the headings in a section (ugly code, could be better, just for example):
```ocaml
let printer =
  let headings = ref [] in
  in
  let print_heading (p: Omd.printer) b (i, md) =
    let f j =
      if i <= j
      then
        (Buffer.add_string b (Printf.sprintf "</section>\n"); false)
      else
        true in
    headings := i::(List.filter f !headings);
    let id = Buffer.create 64 in
    Buffer.add_string b (Printf.sprintf "<section>\n<h%d>" (Bytes.to_string (Buffer.to_bytes id)) i i);
    p.inline p b md;
    Buffer.add_string b (Printf.sprintf "</h%d>" i);
  in
  let finish_print _ b =
    List.iter (fun _ -> Buffer.add_string b (Printf.sprintf "</section>\n")) !headings
  in
  { Omd.default_printer with heading = print_heading; finish = finish_print }

let omd input output =
  let output = open_out_bin output in
  let process ic =
    let md = Omd.of_channel ic in
    output_string output (Omd.to_html ~printer md);
    flush output
  in
  let ic = open_in input in
  match process ic with
  | () -> close_in ic
  | exception e -> close_in_noerr ic; raise e
```

This is more a draft than a final work, so reviews would be really appreciated :)

### DIFF
--- a/src/ast.ml
+++ b/src/ast.ml
@@ -1,13 +1,3 @@
-[@@@warning "-30"]
-
-type list_kind =
-  | Ordered of int * char
-  | Unordered of char
-
-type list_style =
-  | Loose
-  | Tight
-
 type 'a link_def =
   {
     label: 'a;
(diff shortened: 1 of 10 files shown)


src/html.ml
```diff
@@ -1,5 +1,29 @@
 open Ast
 
+type printer =
+  {
+    start: printer          -> Buffer.t -> inline block list                                             -> unit;
```
nojb:
```suggestion
    document: printer          -> Buffer.t -> inline block list                                             -> unit;
```

src/html.ml
```diff
@@ -1,5 +1,29 @@
...
+    finish: printer         -> Buffer.t                                                                  -> unit;
```
nojb: This one does not need to be part of the API. One can always override `document` (i.e. `start` above) to perform some work after the processing of the document is finish.

src/html.ml
```diff
@@ -1,5 +1,29 @@
...
+    block: printer          -> Buffer.t -> inline block                                                  -> unit;
+    paragraph: printer      -> Buffer.t -> inline                                                        -> unit;
+    blockquote: printer     -> Buffer.t -> inline block list                                             -> unit;
+    list: printer           -> Buffer.t -> list_kind * list_style * inline block list list               -> unit;
```
nojb: Since these are functions we could use multiple arguments instead of tuples. Even better, why not define suitable record types and use them both for the types `block` and `inline` and here?

src/html.ml
```diff
@@ -77,128 +101,215 @@ let text_of_inline ast =
...
+  | Hard_break ->
+    p.hard_break p b
+  | Soft_break ->
+    p.soft_break p b
+  | Html body ->
+    p.html p b body
+  | Link (kind, link) ->
+    p.link p b (kind, link)
+  | Ref (kind, md, link) ->
+    p.ref p b (kind, md, link)
```
nojb: It looks like you are not using `ocp-indent`.

clecat: Yes my bad, will do it now

src/html.ml
clecat: Indeed, I kept the tuples without thinking about it, thx

src/html.ml
clecat: Do you want me to use ocaml-format as well ?

nojb: No, I haven't looked into it yet. Let's leave that for later.

clecat:
I havent done the record types idea yet, but the other reviews were taken into account

src/omd.mli
```diff
@@ -55,10 +55,14 @@ type inline =
 type t = inline block list
 (** A markdown document *)
 
+type printer = Html.printer
```
nojb: Shouldn't we re-export the type here (`type printer = Html.printer = {...}`), since we don't expose the `Html` submodule?

nojb:
Where are we with this? If you prefer we could leave the improvement of types (using records) for later.

clecat:
Sorry, I didn't take the time to do that lately, I suppose we could, but it should not take too much time, I will work on it tomorrow

nojb:
Sure, no problem!

clecat:
I did it. However I'm not sure that's what you expected, tell me if you want me to change that

src/ast.ml
```diff
@@ -39,28 +54,45 @@ type link_kind =
   | Img
   | Url
 
-type inline =
+type emph =
+  {
+    kind: emph_kind;
+    style: emph_style;
+    md: inline;
+  }
```
nojb: One possibility is to use different modules to group types for each node type, e.g.:
```
module Emph = struct
  type inline
  type kind = (* definition of type emph_kind *)
  type style = (* emph_style *)
  type 'inline t =
    { style: style;
      kind: kind;
      content: 'inline }
end 
```
Similarly for modules `Link`, `Ref`, `Code_block` and `List`. The type argument to `t` is not very nice but it is necessary to break the recursion.

What do you think?

clecat: Seems good, I will give it a try (sorry for the delay)

nojb: Sure, no worries! Am not 100% this is the best design, but I feel it is worth a try.

clecat: I had no time to do it last week, but it's done now. It looks like it's working properly, and it cleans a few things.

I removed the `type inline`, I did not see the interest of it

clecat:
Ok, it works, I tried using it with mdx and I was able to do what I wanted.
There is place for improvement, but for now, it seems usable

nojb:
> Ok, it works, I tried using it with mdx and I was able to do what I wanted.
> There is place for improvement, but for now, it seems usable

Great! I will read the code more attentively later today and merge. Thanks!

nojb:
Looks good, thanks!

Issue #186, opened 2019-05-24
Allowing attributes extensions parsing


clecat:
Pandoc supports several extensions allowing the declaration of attributes for various 
purposes, among others:
- [Heading attributes](https://pandoc.org/MANUAL.html#extension-header_attributes)
- [Fenced code attributes](https://pandoc.org/MANUAL.html#extension-fenced_code_attributes)
- [Inline code attributes](https://pandoc.org/MANUAL.html#extension-inline_code_attributes)
- [Link attributes](https://pandoc.org/MANUAL.html#extension-link_attributes)
- [Fenced divs attributes](https://pandoc.org/MANUAL.html#extension-fenced_divs)

Their format which seems to be common to all of them consists in curly brackets containing the needed attributes:
> {.idx}

This format however, is not part of the commonmark specs (apparently discussed [here](https://talk.commonmark.org/t/consistent-attribute-syntax/272) since 2014), and therefore is not implemented in omd.

Could it be possible to extend the parsing of omd, so we could have the parsing of those attributes if asked by the user ?

clecat:
For example as an additionnal optionnal string/string list added in the records.

nojb:
Seems reasonable. If we add this we should add it unconditionally, I want to keep different code paths in the parser and the code in general to the absolute minimum.

To add support for attributes you need to modify the parser. To get started you could start by adding attributes to a paragraph-level node. The simplest is probably the case of headings. You will need to

- modify the function https://github.com/ocaml/omd/blob/master/src/parser.ml#L335-L358 in order to recognize optional attributes at the end of the line. Look at the existing parsers for inspiration.

- propagate this data inside the `Parser.Latx_heading` and `Ast.Heading` constructors.

- adapt the printers (at least the HTML one)

- add some tests

clecat:
Okay thank you for the informations, I will give it a look then

avsm:
Thanks for the guidance @nojb!  A number of these are extremely useful for RWO (which motivates @clecat's request).  We also need to decide whether the ocaml.org content will support these attributes well, before we switch that over to a released omd trunk to generate that. (cc @Chris00)

Chris00:
I do not think supporting these attributes would be a problem for ocaml.org.  Now, since they are not standard markdown, they probably should be optional and turned off by default (in the module interface and for the executable).

nojb:
> Now, since they are not standard markdown, they probably should be optional and turned off by default (in the module interface and for the executable).

[...]

PR #187, opened 2019-06-18
Adding attributes parsing

clecat:
First iteration of the addition of the parsing of attributes in omd mentionned [here](https://github.com/ocaml/omd/issues/186).

- [x] Adding parsing of attributes
- [x] Interpret heading attributes
- [x] Interpret fenced code attributes
- [x] Interpret inline code attributes
- [x] Interpret link attributes
- [x] Interpret ref attributes
- [x] ~~Interpret fencer divs attributes~~ (not actually supported in omd)

I only added the parsing of heading attributes, so I can get a review before diving more into the other attributes.


nojb:
Cool!

Currently you are representing the attributes as a single string. Are you planning to parse them further (id, classes, html attributes) or are you planning to leave that to external tools?

clecat:
I wasn't sure about what I could do. I was thinking about returning a list of string separated on spaces at least.
Going further into the parsing may be a good idea, but i wanted your return first, as keeping it as a simple string may be more canonical.

nojb:
> I wasn't sure about what I could do. I was thinking about returning a list of string separated on spaces at least.
> Going further into the parsing may be a good idea, but i wanted your return first, as keeping it as a simple string may be more canonical.

I think parsing it into a record `{id: string option; classes: string list; attributes: (string * string) list}` is a good idea. Anyway you will need to do something along these lines in order to adapt the HTML printer, as the PANDOC syntax is not valid HTML (i.e. `{#foo .bar}` needs to be printed as `<... id="foo" class="bar">`).

clecat:
Okay, I willingly ignored the attribute in the html printer, as I wasnt if we wanted to do this parsing, i'll do that

clecat:
I added the full parsing of attributes:
It is quite simple and does not detect some strange cases which could be considered wrong.
The problem is, I used pandoc as a reference, but it has no specification about what the format of attributes should be, and therefore contains a lot of strange cases I discovered (with backslashes, several dots, equal signs and sharps for example).
This is why I decided to keep it simple and ignore any corner case. I'm not sure if this is the best way, but I think we can come back to that later if it's really a problem

clecat:
One other problem, which we discussed in the original issue, is the breaking chances: By modifying the printer, I broke the cow tests whith headers. We might finally need to make attribute parsing optional (not asking for them would juste ignore the parsing).

clecat:
Or at least, we can always  do the parsing, but make the printer ignore the attributes if not asked

nojb:
> I added the full parsing of attributes:

Great!

> It is quite simple and does not detect some strange cases which could be considered wrong.
> The problem is, I used pandoc as a reference, but it has no specification about what the format of attributes should be, and therefore contains a lot of strange cases I discovered (with backslashes, several dots, equal signs and sharps for example).
> This is why I decided to keep it simple and ignore any corner case. I'm not sure if this is the best way, but I think we can come back to that later if it's really a problem

Sounds good, we can always improve it later.

nojb:
> One other problem, which we discussed in the original issue, is the breaking chances: By modifying the printer, I broke the cow tests whith headers. We might finally need to make attribute parsing optional (not asking for them would juste ignore the parsing).

Sorry, can you remind me how exactly the new stuff breaks the cow tests with headers?

clecat:
If not id attribute is here, the default behaviour of pandoc was to use the content of the heading as the id, so I did the same, which breaks the cow tests asking for no id at all.

Therefore doing it was kinda ugly, and I would be glad to remove the default id, it would make everything cleaner and keep the tests working.

nojb:
> If not id attribute is here, the default behaviour of pandoc was to use the content of the heading as the id, so I did the same, which breaks the cow tests asking for no id at all.
> 
> Therefore doing it was kinda ugly, and I would be glad to remove the default id, it would make everything cleaner and keep the tests working.

I see. I agree we should choose one behaviour and stick to it. Not sure what is more common out there, if pandoc's behaviour (generating the id from the header title) or not generating any id at all.

OK, let us just do whatever is simpler (not generating any id by default) and we can always revisit this later if necessary.

clecat:
So there it is, I think I have done everything we needed and which I mentioned in the original issue. Things might need a little bit of feedbackn, but it seems to work pretty well.

For the parsing of the attributes, I choose a really simple way of parsing the attributes. It might accept too much formats, but pandoc was having way too much inexplicit exceptions.

What do you think of the PR @nojb ?

nojb:
Thanks for your work; am a bit busy at the moment, but will take a look in the next couple of days.

src/ast.ml
```diff
@@ -1,10 +1,20 @@
...
+
 type 'a link_def =
   {
     label: 'a;
     destination: string;
     title: string option;
+    attributes: attributes;
   }
 
+let empty_attributes = {id=None; classes=[]; attributes=[]}
```
nojb: let's make an `Attributes` module with a type `t` and a value `empty`

src/ast.ml
```diff
@@ -1,10 +1,20 @@
...
+    attributes: (string * string) list;
+  }
+
 type 'a link_def =
   {
     label: 'a;
     destination: string;
     title: string option;
+    attributes: attributes;
   }
```
nojb: Can we put this in its own module `Link_def.t` ?

src/block.ml
```diff
@@ -17,7 +17,7 @@ module Pre = struct
     | Rblockquote of t
     | Rlist of Block_list.kind * Block_list.style * bool * int * block list list * t
     | Rparagraph of string list
-    | Rfenced_code of int * int * Code_block.kind * (string * string) * string list
+    | Rfenced_code of int * int * Code_block.kind * (string * string) * string list * attributes
```
nojb: It may be a good idea to use inline records here, but we can leave this for later.

src/html.ml
```diff
@@ -10,52 +10,48 @@ type printer =
...
-        if i < 0 || s'.[i] <> '-' then i
-        else loop (i - 1)
-      in
-      loop (String.length s' - 1)
+          loop keep (i + 1)
+      end
     in
-    (* Trim trailing dashes. *)
-    ids#mangle @@ String.sub s' 0 (last_trailing + 1)
+    loop true (trim 0); Bytes.to_string (Buffer.to_bytes id)
```
nojb: Can use `Buffer.contents` here.

src/html.ml
```diff
@@ -100,6 +96,24 @@ let text_of_inline ast =
...
+  end;
+  if List.length attributes.classes <> 0
+  then
+    begin
+      Buffer.add_string b (Printf.sprintf " class=\"%s\"" (String.concat " " attributes.classes))
+    end;
+  if List.length attributes.attributes <> 0
+  then
+    begin
+      let f (d, v) = Buffer.add_string b (Printf.sprintf " data-%s=\"%s\"" d v) in
```
nojb: Why always prefix attributes with `data-`? This seems a bit ad-hoc.

src/markdown.ml
```diff
@@ -48,6 +48,45 @@ let escape_markdown_characters s =
...
   Buffer.contents b
 
+let print_attributes b a =
+  match a with
+  | {id = None; classes = []; attributes = []} -> ()
+  | _ ->
+    let start = ref false in
+    Printf.bprintf b "{";
+    begin
+      match a.id with
```
nojb: Style: usually we put the `match` in the same line as the `begin` to avoid increasing indentation.

src/markdown.ml
```diff
@@ -59,21 +98,26 @@ let rec inline b = function
   | Emph {kind = Strong; style; content} ->
       let q = match style with Star -> '*' | Underscore -> '_' in
       Printf.bprintf b "%c%c%a%c%c" q q inline content q q
-  | Code (n, c) ->
-      let d = String.make n '`' in
-      Printf.bprintf b "%s%s%s" d c d
+  | Code c ->
```
nojb: Can use pattern matching `Code {level; content; attributes}`

src/parser.ml
```diff
@@ -730,6 +818,33 @@ let is_empty st =
...
+        in
+        junk s;
+        loop s (pos s)
+    | _ ->
+        None
+  in
+  try
+    parse_attributes a
+  with
+    | Invalid_argument _ -> empty_attributes
```
nojb: Style: typically we put the exception pattern in the same line as `with` and the exception handler in the next different line.

src/html.ml
```diff
@@ -10,52 +10,48 @@ type printer =
...
     ref: printer            -> Buffer.t -> inline Ref.t              -> unit;
   }
 
-let _id_of_string ids s =
-  let n = String.length s in
-  let out = Buffer.create 0 in
-  (* Put [s] into [b], replacing non-alphanumeric characters with dashes. *)
-  let rec loop started i =
-    if i = n then ()
+let id_of_string s =
```
nojb: Sorry, I can't see where this is used; and why is `s` of type `bytes` and not `string` ?

src/parser.ml
```diff
@@ -1673,12 +1788,17 @@ let link_reference_definition st =
   if next st <> ':' then raise Fail;
   ws st;
   let destination = link_destination st in
+  let pos = pos st in
+  ws st;
+  let attributes = inline_attribute_string st in
+  if attributes = empty_attributes then
+    set_pos st pos;
```
nojb: This logic of resetting the position should go into `inline_attribute_string`, so that if there are no attributes the position is not changed.

src/parser.ml
nojb: Also, which function raises `Invalid_argument` ?

src/parser.ml
```diff
@@ -332,13 +332,100 @@ let setext_heading s =
...
+    let f acc s =
+      match s.[0] with
+      | '#' -> {acc with id = Some (String.sub s 1 (String.length s - 1))}
+      | '.' -> {acc with classes = (String.sub s 1 (String.length s - 1))::acc.classes}
+      | _ ->
+        let attr = String.split_on_char '=' s in
+        try
+          {acc with attributes = (List.nth attr 0, List.nth attr 1)::acc.attributes}
+        with
+          | Failure _ -> acc
```
nojb: Here it is better to use pattern matching on `attr` rather than `List.nth` and catching the exception.

src/parser.ml
```diff
@@ -332,13 +332,100 @@ let setext_heading s =
...
+        loop' (Sub.tail s)
+    | Some c ->
+        Buffer.add_char buf c;
+        loop (Sub.tail s)
+  in
+  let s', a = loop (ws s) in
+  try
+    s', parse_attributes a
+  with
+    | Invalid_argument _ -> s, empty_attributes
```
nojb: Which function can raise here ?

src/ast.ml
clecat: I agree, should have thought about it before after doing all those modules :)

src/ast.ml
clecat: I just thought the same when moving attributes to his own module, felt a bit off to have it alone, done now

clecat: Just one question, when the type checker does not know the fields for the record creation:
eg:
```
File "src/parser.ml", line 1507, characters 7-16:
1507 |       {label; destination; title = None; attributes = Attributes.empty}
              ^^^^^
Error: Unbound record field label
```
Do you prefer that I force the type of the variable on the function to `Ast.Link_def.t`, or that I use `Ast.Link_def.label instead` of `label` ?

nojb: The usual way is to explicitly qualify the label with the module name, i.e. `Link_def.label`.

clecat: Okay, I'll do that then

src/html.ml
clecat: I was using pandoc as a reference for the behaviour of the attributes as there is no standard. But indeed, it could be better without the automatic data prefix

src/block.ml
clecat: I'll try to do that on another PR

nojb: Sure, but don't worry about it too much, it is not urgent :)

src/parser.ml
clecat: The call to s.[0] can throw it if the string is empty. Actually this is bad, it could make fail attributes which contains too much spaces, I'll just ignore empty strings

nojb: OK, in general one should never catch the exception `Invalid_argument` which indicates programmer error; rather one should check explicitly for the condition that raises it (in this case, an empty string, etc).

src/parser.ml
clecat: Done, and this way to do things would cut the attributes value, if it contained an equal (I only used the first element) so I added a concat

src/parser.ml
clecat: Same as the other one, corrected now

src/html.ml
clecat: This is actually a relic of before my work on omd, I may have used it for something and forgot to delete it afterward

nojb:
Looks good. Do you have any other thing left to do here or can we merge?

clecat:
No I think it's all good, I will have to come back for the inline records, but it will be for another PR

nojb:
Thanks!

Issue #190, opened 2019-08-02
Allowing support for Divs and Spans


clecat:
I'm currently working on replacing pandoc for omd in mdx.
But RWO which uses mdx, is currently making use of an extension of pandoc allowing to generate [divs and spans](https://pandoc.org/MANUAL.html#divs-and-spans) in the html.

Fenced divs can be defined by using a succession of at least 3 semi-columns:
```
::::: {#fenced-div}
Foo
:::::
```
Which would give
```
<div id="fenced-div">
<p>Foo</p>
</div>
```

Spans syntax is close to the syntax of links:
```
[Span]{.class}
```
Which would give
```
<p><span class="class">Span</span></p>
```

Could it be possible to add it's support in omd ?
I would be glad to do it if you agreed with the idea.

clecat:
Also, I would had the support for the fenced divs attributes which I dropped in [this pull request](https://github.com/ocaml/omd/pull/187) as we did not support them for now

nojb:
Sorry, I am not familiar with this functionality. Do I understand correctly that the intention is to make `div`s with markdown block content and similarly `span`s with markdown inline content inside?

clecat:
Yes my bad, I was not really clear, that's exactly the idea

nojb:
It seems to me that the best would be to allow these two kinds of extensions (one for block content and one for inline content) in a generic way and allow the clients of the library to interpret the extensions as they want.

For example we could have `{!foo: <inline>}` and
```
<!foo>
<block content>
</!foo>
```

These would be represented by special nodes in the AST with the tag ("foo") and the content. These nodes would be rejected by the AST printers. It would be the responsibility of the library user to, either 1) rewrite the AST replacing these nodes by non-extension ones, or 2) extend the printers to handle the extension nodes.

The syntax above is just a placeholder, and it needs to have a low likelyhood of being confused with usual markdown (and be easy to parse).

What do you think?

A possible approach would be:
1) add the new extension nodes to the AST, rejecting them if they reach the printers
2) define a generic interface to the printers that allows overloading the treatment of extension nodes (an easy way is to use a record in "open recursion" style)
3) decide a syntax and implement a parser for it

clecat:
I'm okay with that, I was going to ask you for a way to have extensions.
I'm just not sure about what you mean with "decide a syntax and implement a parser".
Do you mean that:
- a user needs a parser and a printer for his extension
- or he needs only the printer, and the extensions follows a common syntax

[...]

PR #191, opened 2019-08-13
Added implementation of inline tags and tag blocks

clecat:
This pull request contains the implementation of the tag extension system discussed [here](https://github.com/ocaml/omd/issues/190):

It allows the user to use self defined printers by using a new syntax:
For inline tags:
```
  This is an {!foo: inline tag}.
```
For tag blocks:
```
{!bar:
  This is a tag block
}
```

The html tag printers will be called with an object containing:
- the string of the tag (in the examples, it will be `bar`)
- in the case of an inline tag, the text (`inline tag` for the example)
- in the case of a tag block, the list of the blocks contained in the tag (here, a Paragraph).
- and finally the attributes

Nesting blocks inside a tag block is possible !

This implementation may need some corrections, I'm listening to all your comments and modifications requests :)

### DIFF
--- a/src/parser.ml
+++ b/src/parser.ml
@@ -271,6 +271,7 @@ type t =
   | Llist_item of Block_list.kind * int * Sub.t
   | Lparagraph
   | Ltag of string * Attributes.t
+  | Lendtag
 
 let sp3 s =
   match Sub.head s with
@@ -783,7 +784,7 @@ let tag s =
                   s, Attributes.empty
             in
             let string_tag = Buffer.contents tag in
-            let string_tag = if string_tag = "" then "none" else string_tag in
+            let string_tag = if string_tag = "" then raise Fail else string_tag in
             Ltag (string_tag, a)
         | Some (' ' | '\t' | '\010'..'\013') | None ->
             raise Fail
@@ -827,7 +828,7 @@ let parse s0 =
   | Some '{' ->
       tag s
   | Some '}' ->
-      Ltag ("", Attributes.empty)
+      Lendtag
   | Some _ ->
       (blank ||| indented_code ind) s
   | None ->
--- a/src/parser.ml
+++ b/src/parser.ml
@@ -270,6 +270,7 @@ type t =
   | Lhtml of bool * html_kind
   | Llist_item of Block_list.kind * int * Sub.t
   | Lparagraph
+  | Ltag of string * Attributes.t
 
 let sp3 s =
   match Sub.head s with
@@ -767,6 +768,33 @@ let blank s =
   if not (is_empty s) then raise Fail;
   Lempty
 
+let tag s =
+  match Sub.heads 2 s with
+  | ['{'; '!'] ->
+      let tag = Buffer.create 17 in
+      let rec loop s =
+        match Sub.head s with
+        | Some ':' ->
+            let _, a =
+              match Sub.head ~rev:() s with
+              | Some '}' ->
+                  attribute_string s
+              | _ ->
+                  s, Attributes.empty
+            in
+            let string_tag = Buffer.contents tag in
+            let string_tag = if string_tag = "" then "none" else string_tag in
+            Ltag (string_tag, a)
+        | Some (' ' | '\t' | '\010'..'\013') | None ->
+            raise Fail
+        | Some c ->
+            Buffer.add_char tag c;
+            loop (Sub.tail s)
+      in
+      loop (Sub.tails 2 s)
+  | _ ->
+      raise Fail
+
 let indented_code ind s =
   if indent s + ind < 4 then raise Fail;
   Lindented_code (Sub.offset (4 - ind) s)
@@ -796,6 +824,10 @@ let parse s0 =
       unordered_list_item ind s
   | Some ('0'..'9') ->
       ordered_list_item ind s
+  | Some '{' ->
+      tag s
+  | Some '}' ->
+      Ltag ("", Attributes.empty)
   | Some _ ->
       (blank ||| indented_code ind) s
   | None ->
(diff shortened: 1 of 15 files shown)


nojb:
Cool, am a bit busy at the moment, but will take a look at this over the weekend. Thanks!

clecat:
No problem, no need to rush :) Thanks for giving a look

XVilka:
By the way, it is easier to send pull requests from a separate branch rather than from master - this way you can work on many different pull requests in parallel and rebase/merge/cherry-pick when required.

clecat:
Thanks for the information @XVilka, I'm aware of that dont worry, and I do it from time to time, but laziness (and more often just omission) makes me push from master :p

src/sexp.ml
```diff
@@ -58,6 +60,8 @@ let rec block = function
       List [Atom "html"; Atom s]
   | Link_def {label; destination; _} ->
       List [Atom "link-def"; Atom label; Atom destination]
+  | Tag_block {tag; content; _} ->
+      List [Atom "tag"; Atom tag; List (Atom "list-item" :: List.map block content)]
```
nojb: Why is there a `list-item` here ? Shouldn't this be
```
List [Atom "tag"; Atom tag; List (List.map block content)]
```

src/parser.ml
```diff
@@ -796,6 +824,10 @@ let parse s0 =
       unordered_list_item ind s
   | Some ('0'..'9') ->
       ordered_list_item ind s
+  | Some '{' ->
+      tag s
+  | Some '}' ->
```
nojb: Why is this necessary (returning `Ltag` when seeing `}`) ?

src/parser.ml
```diff
@@ -767,6 +768,33 @@ let blank s =
...
+        | Some ':' ->
+            let _, a =
+              match Sub.head ~rev:() s with
+              | Some '}' ->
+                  attribute_string s
+              | _ ->
+                  s, Attributes.empty
+            in
+            let string_tag = Buffer.contents tag in
+            let string_tag = if string_tag = "" then "none" else string_tag in
```
nojb: I don't think we should return `"none"` if the tag string is empty. In fact, we should probably disallow empty tags. What do you think?

tests/tag/uppercase.ml
```diff
@@ -0,0 +1,88 @@
...
+      | "capitalize" ->
+        let f i block =
+          p.block {p with text = print_upper_text; code = print_upper_code} b block;
+          if i < List.length t.content - 1 then
+            Buffer.add_char b '\n'
+        in
+        List.iteri f t.content
+      | _ -> Omd.default_printer.tag_block p b t
+    in
+    {Omd.default_printer with tag = print_tag; tag_block = print_tag_block}
```
nojb: Style: use punning whenever you can, by calling functions `tag`, `tag_block`, etc.

src/parser.ml
nojb: Ah OK I understand, it is to signal the end of a tagged block. I would use `Lendtag` (to be defined).

src/sexp.ml
clecat: Okay my bad, I never use sexp before and simply refered to the use above, I'll take that into note

src/parser.ml
clecat: Yes, better idea, I wanted to have a minimum of element in the variant, I'll change that

src/parser.ml
clecat: Yes, better idea, thank you

tests/tag/uppercase.ml
clecat: I corrected that and changed it in the html.ml file too: tell me if it's better now

clecat:
About the tag syntax, I did not really think about any other, this one is good to me, but I dont have any preferences, so if you think another one could be better, I'll change it

clecat:
Maybe we should change the syntax for something more common: We could copy que way code blocks are written with '~', something like:
```
~~~foo
bar
~~~
```
For blocks and
~~~
This is ~foo bar~
~~~
For inline tags (we may need a better way to specify the tag name for them)

What do you think of it ? people might be more comfortable with this as it looks like code blocks

clecat:
My bad, '~' are already used for code blocks too, but if we find another good character, it could be a good idea

pw374:
> Maybe we should change the syntax for something more common: We could copy que way code blocks are written with '~', something like:
> 
> ```
> ~~~foo
> bar
> ~~~
> ```
> 
> For blocks and
> 
> ```
> This is ~foo bar~
> ```
> 
> For inline tags (we may need a better way to specify the tag name for them)
> 
> What do you think of it ? people might be more comfortable with this as it looks like code blocks

> My bad, '~' are already used for code blocks too, but if we find another good character, it could be a good idea

Perhaps use `+` (it already has a meaning for itemized lists but it should be ambiguous in your case)

clecat:
Let's go for that

clecat:
Hello, I changed the syntax so it uses fenced code format instead.
I mostly copied the behaviour of fenced code so I would add a minimum of errors
Tell me what do you think of the current code (parsers may need some cleaning)

clecat:
Any thoughts about it ? :D

nojb:
> Any thoughts about it ? :D

Sorry for the delay. I will take a look in the next few days. Thanks!

clecat:
No problem, I don't want to steal your time :p

clecat:
Got any time to give it a look ?

nojb:
> Got any time to give it a look ?

Hello @clecat, very sorry for the delay in looking at this, I have been swamped with other things. I will definitely look at this before monday. Thanks for the patience!

nojb:
And yes, the parser part is tricky. I would like to clean that up and switch to something more understandable (maybe parser combinators?).

clecat:
Thanks a lot for your review :D

I agree with you about the special case of the inline tags, it annoys me a lot, I should try to fing a best way to do that (maybe `+foo+this is the bar text+` in order to keep the tag clearly separated)

And I agree with the parser part too, I've finally understood how it works but a little bit of rework refactor or redesign could be nice, if I have the opportunity and you want me too, I could give it a look for another PR.

But for now do you agree with the PR ? Do you want me to change anything ? (I've also 2 smalls PRs backed for tables and definition list which could be merged before I start looking at the parser refactoring)

nojb:
Seems OK except for the choice of syntax. I will merge this and let you take the lead in finding a better alternative later on. Refactoring the parser would be very welcome! Thanks!

clecat:
No problem, I'll try to change it when I can

Just to be sure, what is bothering you is the inline special syntax ? Or the whole syntax

nojb:
> Just to be sure, what is bothering you is the inline special syntax ? Or the whole syntax ?

In principle the inline syntax is a problematic. But more generally I think that we need to have some experience with this feature to know if this syntax is "right" or not.

Issue #188, opened 2019-07-09
Feature request: handle extended markdown


rjbou:
`omd` is used by opam2web, to generate opam.ocaml.org pages. Table rendering is now handled by omd as a plain text. It is one of the feature of the [extended syntax](https://www.markdownguide.org/extended-syntax).

I can share an image example if needed.

nojb:
I agree that tables are useful to have. GFM-style tables are a suitably design, I think. Now all we need is an implementation.

avsm:
This would be useful for RWO as well; we could do this as part of that effort once the RWO build migrates to omd master (cc @clecat )

clecat:
It seems nice to do, I will see if I can do that as a secondary effect of my work on the migration

XVilka:
To date, most of the Markdown parsers and producers support at least some of the extensions, so having them indeed is very useful.

nojb:
We are planning to have them. But for the initial release of version 2 (which we are working to have out the door as soon as possible), we are going to focus on having strict commonmark compatibility. Extensions will come after that. Thanks!

nojb:
Moved this one to #205

PR #192, opened 2019-10-15
Implementation of definition lists

clecat:
This pull request adds the implementation of definition lists in markdown files.
This feature is not part of the commonmark specification, but was requested in [this issue](https://github.com/ocaml/omd/issues/188) as being a part of the [extended markdown syntax](https://www.markdownguide.org/extended-syntax).

The syntax of definition lists is pretty easy:
```
First Term
: This is the definition of the first term.

Second Term
: This is one definition of the second term.
: This is another definition of the second term.
```

The parsing is dumb, but works actually pretty well:
Lines of raw text beginning are considered as `Ldef_list` tokens, and if during the process of the blocks, they are preceded by one single line of raw text or an already created definition list block, a definition list block is created (or fed with the new line).

### DIFF
--- a/src/block.ml
+++ b/src/block.ml
@@ -20,6 +20,7 @@ module Pre = struct
     | Rfenced_code of int * int * Code_block.kind * (string * string) * string list * Attributes.t
     | Rindented_code of string list
     | Rhtml of Parser.html_kind * string list
+    | Rdef_list of string * string list
     | Rtag of int * int * string * t * Attributes.t
     | Rempty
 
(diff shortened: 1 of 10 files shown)


tests/def_list/def_list.md
```diff
@@ -0,0 +1,8 @@
+First Term
+: This is the definition of the first term.
+
+Second Term
+: This is one definition of the second term.
```
nojb: Can you have multiline definitions?

src/block.ml
```diff
@@ -125,6 +137,10 @@ module Pre = struct
             s
         in
         {blocks; next = Rfenced_code (ind, num, q, info, Sub.to_string s :: lines, a)}
+    | Rdef_list (term, d::defs), Lparagraph ->
+        {blocks; next = Rdef_list (term, (d ^ "\n" ^ (Sub.to_string s))::defs)}
```
nojb: Can you say a few words about this case?

tests/def_list/def_list.md
clecat: Ho yes, my bad I forgot to add a test, and this is actually what does the case you mentionned below

src/block.ml
clecat: This case is what adds the support for multiline definitions:
If we find a paragraph line after an Rdef_list, we append it to the last definition.
I'll add an example

clecat:
Any additional thoughts ?

clecat:
Do you have any other thoughts about this PR ? Or can it be merged ?

nojb:
Looks good, thanks.

PR #195, opened 2019-11-15
Iterator and Mapper on the Ast

madroach:
Hi,

I needed a Mapper on the markdown Ast.
After implementing this it was only little work to also do a generic iterator.
I could easily rewrite the Html.default_printer using this iterator, but since the iterator provides no "context" argument, I had to wrap the html printer into a ``to_string`` closure instead of explicitely passing a ``Buffer.t`` around.
A nice side-effect is that one can now easily print directly to a channel without first filling a buffer.
I'm not sure whether the html part is an improvement, but at least it was a test case for the iterator. That's why I split off the html part into a separate commit. I would love to see the first commit merged, but am unsure about the second.

### DIFF
--- a/src/ast.ml
+++ b/src/ast.ml
@@ -182,3 +182,228 @@ let defs ast =
     | Link_def def -> def :: acc
   in
   List.rev (List.fold_left loop [] ast)
+
+module Mapper =
+struct
+  type t =
+    { document:       t -> inline block list         -> inline block list
+    ; blocks:         t -> inline block list         -> inline block list
+    ; attributes:     t -> Attributes.t              -> Attributes.t
+    ; block:          t -> inline block              -> inline block
+    ; paragraph:      t -> inline                    -> inline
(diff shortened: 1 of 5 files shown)


madroach:
Ok, I could not resist and now also tried how it works out with context added to the Iterator. So without context was a5673bd, with context is 5cdad37. The latter is nicer in my opinion.

nojb:
@madroach Thanks for your contribution! I will take a look as soon as possible.

nojb:
I took a look. I don't feel the iterator would be super-useful as the same can be achieved with just pattern matching. The mapper could be a bit more useful, but I changed the main ast type somewhat in preparation for 2.0 so it doesn't apply anymore as-is. In any case, I don't feel the mapper functionality would be useful to have *in* the library, as it is something that can be written easily by the user as needed. So I'm closing this but feel free to reopen if you need to.

PR #196, opened 2019-11-18
print attributes on Tag and Tag_block

madroach:
Hi,

I'm looking for a way to add a class to a paragraph. I have the idea that ``Tag_block`` is supposed to do something like that, but could not figure out what exactly hte ``Tag`` and ``Tag_block`` entities are supposed to do.
I figured out that ``Tag_block`` parses stuff like 
``` markdown
+++ div {.style1 #greeting key=value}
Hallo, Welt!
+++
```
But what is it supposed to render to in html? Is this?
``` html
<div class=style1 id=greeting key=value>
Hallo, Welt!
</div>
```
Because that's what I'm looking for.

### DIFF
--- a/src/ast.ml
+++ b/src/ast.ml
@@ -182,3 +182,228 @@ let defs ast =
     | Link_def def -> def :: acc
   in
   List.rev (List.fold_left loop [] ast)
+
+module Mapper =
+struct
+  type t =
+    { document:       t -> inline block list         -> inline block list
+    ; blocks:         t -> inline block list         -> inline block list
+    ; attributes:     t -> Attributes.t              -> Attributes.t
+    ; block:          t -> inline block              -> inline block
+    ; paragraph:      t -> inline                    -> inline
+    ; blockquote:     t -> inline block list         -> inline block list
+    ; list:           t -> inline block Block_list.t -> inline block Block_list.t
+    ; list_item:      t -> inline block list         -> inline block list
+    ; code_block:     t -> Code_block.t              -> Code_block.t
+    ; thematic_break: t -> unit                      -> unit
+    ; html_block:     t -> string                    -> string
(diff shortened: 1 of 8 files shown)


clecat:
Hello,
About the `Tag` and `Tag_block`, I should try to add some documentation, sorry for the inconvenience.
The goal of the tag block is not to add a class, but to allow the user to do anything he wants, you only have to overwrite the `tag_block` function of the printer, with a function, verifying that your tag is the good one, and then print whatever you want, even, calling back other functions of the printer.

And if you want to make it print a class, you can do it, but it is not it's only goal, you can look at the example `tag_uppercase` to have a more usefull example

clecat:
For what you want, you could go with this code for example:
```
    let tag_block (p: Omd.printer) b t =
      match t.Omd.Tag_block.tag with
      | "div" ->
        Buffer.add_string b "<div";
        p.attributes p b c.Code_block.attributes;
        Buffer.add_string b ">";
        List.iter (p.block p b) t.content;
        Buffer.add_string b "</div>";
      | _ -> Omd.html_printer.tag_block p b t
```

madroach:
Ok. got it. You still might want to cherry-pick 1fd8a1f963..c27416ddd

nojb:
Thanks, we may cherry-pick some of the commits, but they need some adaptations as the codebase has been changing in preparation for 2.0. Furthermore we are discussing whether we will keep the "Tag" stuff for the initial release of 2.0 or whether we will delay adding extension to later. I will update here once a decision is taken.

nojb:
For info, we decided to remove the "tag" stuff from the AST and instead stick to extensions which exist in other comparable projects (pandoc, etc). It did not seem like a good direction to further increase an already fragmented landscape of markdown extensions. On that note, I am closing this PR. Thanks and apologies for having wasted your efforts!

PR #197, opened 2019-11-19
[WIP] Add latex printer

clecat:
Hello,

This PR's goal is to add a printer for `latex` output.
I'm not really sure this is something we want, but it could be useful, tell me what you think of it.

For now the implementations are empty, as I'm no latex expert, I'll try to do my best in order to fill them, but advice are always welcome.

To do:
- [x] document
- [x] attributes
- [x] block
- [x] paragraph
- [x] blockquote
- [x] list
- [x] code_block
- [x] thematic_break
- [x] html_block
- [x] heading
- [x] def_list
- [x] tag_block
- [x] inline
- [x] concat
- [x] text
- [x] emph
- [x] code
- [x] hard_break
- [x] soft_break
- [x] html
- [x] link
- [x] ref
- [x] tag

### DIFF
--- a/bin/main.ml
+++ b/bin/main.ml
@@ -1,6 +1,7 @@
 type format =
   | Sexp
   | Html
+  | Latex
   | Markdown
   | Text
 
@@ -9,6 +10,7 @@ let fmt = ref Html
 let convert ast =
   match !fmt with
   | Html -> Omd.to_html ast
+  | Latex -> Omd.to_latex ast
   | Sexp -> Omd.to_sexp ast
   | Markdown | Text -> assert false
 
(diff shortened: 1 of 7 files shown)


clecat:
I added in the last commit a first implementation of the printer, using the output of pandoc and what seemed to me to be a common way of doing things.

However there may be several problems with escaping for example or simply wrong /not optimal implementations.

Reviews from someone with more knowledge of `latex` would be really appreciated.

nojb:
This looks useful. It doesn't apply anymore to `master` because I got rid of the `printer` type in preparation for to 2.0. But let's discuss on Tuesday about this as well, and whether we want to have this in for 2.0 or we leave it for after.

nojb:
As discussed: this is delayed until after the release.

shonfeder:
We'll revisit this once we've settled the current AST. If @clecat is interested in updating the PR to work with all the changes that have been added subsequently, we'll provide whatever support we can. If not, I'll plan to do the integration work myself.

Thanks for your patience here!

PR #203, opened 2020-06-19
Introduce an intermediate representation for HTML

nojb:
This PR does several things:

- Introduce an intermediate representation for HTML, greatly simplifying the HTML backend
- Get rid of the standalone text backend and instead do a simple HTML-to-text conversion
- Get rid of the markdown "backend" (it wasn't working and of no obvious utility, perhaps except for testing)

As a consequence of the above, I am able to remove completely the "printer" data structure, which was kind of badly thought out.

To be discussed: fate of extensions (tags and attributes). I am leaning to removing/disabling it for a first release of 2.0.0 and aim for full and strict Commonmark compatibility instead. After that, we could look into adding a proper API for extension. 

Note that several tests are failing, but they are due to small whitespace mismatches. I am looking into it.

@clecat


nojb:
@clecat I would like to discuss what to do with the extensions you added. I'll be reaching out offline to chat about this.

PR #204, opened 2020-06-20
Refresh testsuite, upgrade spec to 0.29, switch to dune 2

nojb:
This PR:

- Improves the testsuite: instead of committing the test cases from the official spec, they are extracted each time from the spec. This makes it much easier to update the tests when the spec changes.

- Switch to dune 2, automatically generate OPAM file, etc.

- Upgrade the spec to 0.29 (it is now committed to `tests/spec.txt`).


nojb:
Testsuite failures are expected, to be fixed.

Issue #82, opened 2013-11-28
TyXML


kit-ty-kate:
Is it possible to have an optional module depending on TyXML that have a function to transforms an Omd.t into a TyXML type (Html5_sigs.T.elt) ? It would facilitate the integration with projects that use ocsigen.

agarwal:
@jpdeplaix @darioteixeira Curious if either of you have an update on this. If there's no clear solution still, I might consider implementing something.

darioteixeira:
@agarwal: In trunk, Lambdoc's support for Markdown (via OMD) is much better now, though there are still some issues to resolve before a 1.0 release (if you want to try this route note that Lambdoc's trunk depends on Tyxml's trunk).  As for direct support for Tyxml in OMD, I don't know of any developments.

agarwal:
@darioteixeira Thanks. I'll take a look.

shepard8:
There is still work to be done, but I have a first "quick and dirty" solution to this issue (see above commit). 

The more complex constructors in the `Omd.element` type are not or badly documented. It would help me if someone could explain to me what are
  * Paragraph and how to enable it in `of_string`;
  * Ulp, Olp (I read somewhere that paragraphs are implicit, but the LI tag accepts only phrasing content in the HTML specification);
  * NL
  * Ref
  * Img_ref
  * X

Besides implementing these correctly, I also need to provide the code inside a functor, so that any TyXML module can use it. I also plan to provide submodules D and F compatible with Eliom_content.Html.{D,F}.

Any comment on that would be welcome. Am I heading in the right direction?

jpdeplaix:
Maybe @Drup can answer to some of these questions

Drup:
I would be happy to answer tyxml questions, but I feel the implementation above does a decent job at trying to turn the loosely-typed omd AST into tyxml. The questions seems mostly omd-specific.

shonfeder:
I've encountered a need for this as well, so thought I'd revive the old discussion. 

@shepard8 are you still interested in pushing this through? If so, I'd be happy to see if I could help figure out how we should treat the listed constructors. (It not, I'd be willing to pick this issue up.)

shepard8:
Hello!

I have no time for this at the moment, feel free to continue (or restart from scratch, I don't know your plans :-) ) the work on this issue.

shonfeder:
Ok! Thanks @shepard8. I should be able to take a crack at this in the next week or two.

nojb:
[...]

PR #211, opened 2020-06-28
Omd_tyxml

shonfeder:
This is just at an early exploratory stage, to explore out the space and be sure I'm headed in the right direction, before I actually set out in earnest.

Closes: #82


src/omd.ml
```diff
@@ -1,9 +1,12 @@
 module Pre = Block.Pre
 
+module Html = Html
+
 include Ast
 
 type doc = block list
 
+
```
nojb, line 9: Spurious diff.

src/omd.mli
```diff
@@ -59,11 +59,14 @@ and block_desc =
...
 
-type doc = block list
+type doc = Ast.block list
 (** A markdown document *)
 
+module Html = Html
+
 val of_channel: in_channel -> doc
 
+
```
nojb, line 69: Spurious diff.

src/omd.mli
```diff
@@ -59,11 +59,14 @@ and block_desc =
   | Link_def of string link_def
   | Definition_list of def_elt list
 
-type doc = block list
+type doc = Ast.block list
```
nojb, line 62: Why this change?

src/omd.mli
```diff
@@ -59,11 +59,14 @@ and block_desc =
...
 (** A markdown document *)
 
+module Html = Html
```
nojb, line 65: If you need to access the definitions from another module, we should repeat the definitions here. The existence the `Html` submodule is a an implementation detail in my mind which may contain other stuff apart from that that we want to expose. The `Omd` interface is the public API of the library and so should stand on its own.

tests/dune.inc
```diff
@@ -4546,7 +4546,7 @@
  (alias spec-649)
  (action (diff spec-649.html spec-649.html.new)))
 (alias
- (name runtest)
+ (name test-specs)
```
nojb, line 4549: Why this change?

tests/omd_tyxml.ml
```diff
@@ -0,0 +1,51 @@
...
+<html xmlns="http://www.w3.org/1999/xhtml"><head><title></title></head><body></body></html>|};
+
+  test "a plain line"
+    ~actual:(html_of_string "a plain line")
+    ~expected:{|<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml"><head><title></title></head><body><p>a plain line</p></body></html>|};
+
+  test "emphasized text"
+    ~actual:(html_of_string "some *emphasized* text")
+    ~expected:{||}
```
nojb, line 29: We typically use `dune` `(diff ...)` rules for this kind of tests which gives us `promote` functionality for free.

dune-project
```diff
@@ -21,3 +21,13 @@ Additionally, OMD implements a few Github markdown features, an
 extension mechanism, and some other features. Note that the opam
 package installs both the OMD library and the command line tool `omd`.")
  (tags (org:ocamllabs org:mirage)))
+
+(package
+ (name omd-tyxml)
+ (synopsis "A library to convert OMD's markdown representation to Tyxml")
+ (description
+  "This optional library enables user of OMD to convert values of type Omd.t,
```
nojb, line 29: I believe our type is called `Omd.doc`.

dune-project
```diff
@@ -21,3 +21,13 @@ Additionally, OMD implements a few Github markdown features, an
...
+representing parsed markdown, into values of type Tyxml.t, which provides
```
nojb, line 30: I don't think there is a type `Tyxml.t`.

shonfeder:
Thanks very much for the initial feedback, @nojb! And sorry for the *very* rough state! As I noted, I only meant to rough in some broad strokes to get direction on a few major points, and your replies here -- and on the originating issue -- have been super helpful and informative in this direction.

shonfeder (commented):
AH! I forgot to submit these comments. They help make more sense

tests/dune.inc
shonfeder: I assume we want `dune (test|runtest)` to run all the tests for the packages bundled in this repository, so we don't want `runtest` to only be assigned to this this set of tests. Thus, this suite is renamed, so we I could define the special `runtest` alias to point to both suites.

src/omd.mli
shonfeder: Just needed to expose the fact that `block` is the same as `Ast.block` to the type checker, so that it could recognize it's occurrence in the reexported `Html` module.

However, since the plan is to just build the `Tyxml` off of the `Omd.doc`, this will be a mute point.

tests/dune.inc
nojb: This can be done without renaming this alias: `alias` stanzas are not "exclusive", there may be many `alias` stanzas attaching dependencies to the same alias.

nojb:
@shonfeder let me know when the code is ready for another review pass. Thanks!

shonfeder:
I've still got some cleanup to do here, but the basic functionality is in place and, afaik, all the tests should be passing *in spirit*. I'm finding the tests quite testy, due to trivial white space differences and arbitrary formatting differences (some end with new lines, some done, one way of generation produces breaks after certain tags, another way doesn't, etc).

I pulled in lambdasoup as a testing dep hoping to fully normalize the HTML so we wouldn't have to fuss with string munging, but it's not working as well as expected. In fact, for some reason some (but not all!) of the HTML test cases are ending up with duplicated tags, causing spurious test failures. I have no idea what's causing these, but I'll look with fresh eyes later in the week.

nojb:
> I've still got some cleanup to do here, but the basic functionality is in place and, afaik, all the tests should be passing _in spirit_. I'm finding the tests quite testy, due to trivial white space differences and arbitrary formatting differences (some end with new lines, some done, one way of generation produces breaks after certain tags, another way doesn't, etc).

One easy way around this is to pretty-print the expected HTML output found in the test files. The output is already being extracted to generate the `.html` files for each test. It is just a matter of inserting a pretty print call before storing the `.html` field in `parse_test_spec` function in `extract_tests.ml`. Then you won't need to do any normalization (apart from the pretty printing).

I was resisting to do this because when pretty-printing there is always the chance that a bug may squeak in. But if it is getting too hard to nail the spec formatting on the head, I think this is a valid way to go.

nojb:
> I pulled in lambdasoup as a testing dep hoping to fully normalize the HTML so we wouldn't have to fuss with string munging, but it's not working as well as expected. In fact, for some reason some (but not all!) of the HTML test cases are ending up with duplicated tags, causing spurious test failures. I have no idea what's causing these, but I'll look with fresh eyes later in the week.

Sounds good, let me know when it is ready for another review pass. Thanks!

omd_tyxml/omd_tyxml.ml
```diff
@@ -0,0 +1,140 @@
...
+  | Link l     -> [of_link attrs l]
+  | Soft_break -> Html.[txt "\n"]
+  | Text t     -> Html.[txt t]
+
+and of_link_label ({il_attributes; il_desc} as il : Omd.inline) =
+  let _attr = il_attributes in
+  match il_desc with
+  | Code _ | Text _   | Concat _
+  | Emph _ | Strong _ | Image _ -> List.map Html.Unsafe.coerce_elt (of_inline il)
+  | _  -> raise (Failure "TODO invalid link label")
```
Drup, line 51: Instead of doing that, I suggest you simply create a function that does a "striped-down" translation, ignoring all the pieces that shouldn't appear in links. It duplicates the code of `of_inline` a little, but it's cleaner and it never fails.

omd_tyxml/omd_tyxml.ml
```diff
@@ -0,0 +1,140 @@
...
+
+and of_link attrs (l : Omd.link) =
+  let escaped_url = Omd.Internal.escape_uri l.destination in
+  let attrs =
+    (Html.a_href escaped_url :: attrs) @ (Option.map Html.a_title l.title |> Option.to_list)
```
Drup, line 56: For all these operations, I suggest creating a `cons_opt` function/operator. :)

omd_tyxml/omd_tyxml.ml
```diff
@@ -0,0 +1,140 @@
...
+  let h =
+    let open Html in
+    match n with
+    | 1 -> h1
+    | 2 -> h2
+    | 3 -> h3
+    | 4 -> h4
+    | 5 -> h5
+    | 6 -> h6
+    | m -> raise (Invalid_markdown (Printf.sprintf "heading number %d" m))
```
Drup, line 76: If that kind of markdown is truly malformed, the internal representation shouldn't allow it. Otherwise, find some HTML to represent it (using paragraphs). Don't fail on the output to HTML.

omd_tyxml/omd_tyxml.ml
```diff
@@ -0,0 +1,140 @@
...
+  | Thematic_break             -> Html.hr ()
+  | Heading (n, content)       -> of_heading n attrs content
+  | Code_block (src, code)     -> of_code_block src attrs code
+  | Html_block html            -> Html.Unsafe.data html
+  | Definition_list content    -> of_definition_list content
+
+and of_list typ spacing items =
+  let of_list_block (bl : Omd.block) =
+    match bl.bl_desc, spacing with
+    | Paragraph il, Tight -> of_inline il |> List.map Html.Unsafe.coerce_elt
```
Drup, line 103: You should not need to `coerce` here. You might need to think carefully about the kind of stuff you are constructing, and type-annotate in consequence.

omd_tyxml/omd_tyxml.ml
```diff
@@ -0,0 +1,140 @@
...
+let rec of_inline ({il_attributes; il_desc} : Omd.inline) =
+  let attrs = of_omd_attributes il_attributes in
+  match il_desc with
+  | Code c     -> of_code attrs c
+  | Concat ls  -> List.concat_map of_inline ls
+  | Emph e     -> Html.[em ~a:[] (of_inline e)]
+  | Strong s   -> Html.[strong ~a:[] (of_inline s)]
+  | Hard_break -> Html.[br ~a:[] ()]
+  (* TODO Add option for verified html ?*)
+  | Html raw   -> Html.Unsafe.[data raw]
```
Drup, line 40: This is "correct" as long as the HTML is textual. As soon as you start using the non-textual Tyxml backends (for instance, Tyxml_js`), this is broken and you have to do more work.

Long term, I think it would be beneficial to provide a functor over the HTML module to adapt to different Tyxml instantiations (and provide a pre-applied instance for text!). The "right" solution is then to decode the HTML with lambdasoup and build actual tyxml trees. Then you can build DOM trees from markdown directly, without going through text.

shonfeder:
> I was resisting to do this because when pretty-printing there is always the chance that a bug may squeak in. But if it is getting too hard to nail the spec formatting on the head, I think this is a valid way to go.

Your fears have been validated, @nojb! I've been going a bit crazy trying to figure out why there is weird duplicate html in some of the generated test fragments (causing the parsing checks to fail). Turns out, a bug indeed squeak in :)

```
utop # let html = {|<p><a href="foo\
bar"></p>
|};;
val html : string = "<p><a href=\"foo\\\nbar\"></p>\n"

utop # print_endline html;;
<p><a href="foo\
bar"></p>

- : unit = ()
utop # Soup.(parse html |> pretty_print) |> print_endline;;
<p>
 <a href="foo\
bar"></a>
</p>
<a href="foo\
bar">
</a>
- : unit = ()
````

**Comment by Drup** [comment]
Well, that's because your HTML is incorrect (unclosed `a` tag that is not converted to a standalone element), which triggers recovery in lambdasoup.

**Comment by shonfeder** [comment]
Oof! I nearly forgot about this! Life has came at us fast this year :grimacing: 

I don't have time to dig into this again this weekend, but I hope to revisit the next weekend. IIRC, this isn't far off really, and thanks to Drup's feedback it should be easily to make the suggested improvements and push through the final bit.

File: omd_tyxml/omd_tyxml.ml (hunk 13 above)
Comment (shonfeder) [reply omd_tyxml/omd_tyxml.ml:46]: I'm not sure whether we want this generalization as part of the current PR, or as a followup, but I agree we should target it in the long run.

I think we first need to settle settle whether or not we can replace the bespoke html generator with Tyxml.

File: tests/common.ml [hunk 14]
```diff
@@ -0,0 +1,4 @@
+let normalize_html s =
```
shonfeder, line 1: We won't need this if we decide to go with one package.

tests/dune
```diff
@@ -1,8 +1,15 @@
 (executable
  (name extract_tests)
- (libraries str)
+ (libraries str common)
  (modules extract_tests))
 
+;  Code shared between various parts of the testing apartus
```
shonfeder, line 6: Ditto re: only needing this we go the two packages route.

shonfeder (commented):
I think this ready for review, and for us to decide whether we want the Tyxml backend to replace the bespoke html generation or not! 

Thanks @nojb and @Drup for the patience, and sorry for the verrrrry looong turnaround here.

shonfeder:
None of this has any purchase in my working memory any more, the AST has changed, and there are numerous conflicts. So I'm gonna close this and will consider opening separately at some point.

Issue #210, opened 2020-06-27
Lower OCaml requirement


jfrolich:
Can we have some conditional compilation to make this library work under `4.04.2`. I think it would be great to make this accessible to most projects. I am dealing with a project that need to support `4.04` as the minimum version.

I did make it compile on 4.04 by removing `Fun.protect` and adding a `Compat` module for `Buffer.add_utf_8_uchar`, `List.find_opt`, `Uchar.rep`.

nojb:
Hello, and thanks for opening this issue. We are currently preparing a new release and we will look into making it compatible with 4.04.

Just out of curiosity, can you say a few words why you are required to support 4.04 ? It is a rather ancient version of OCaml :)

jfrolich:
We are planning to support (extended) markdown to ocamldoc compilation in the Reason toolchain, the Reason toolchain needs to support 4.04 for now.

nojb:
OK, thanks.

jfrolich:
This is why I am trying to create a printer to ocamldoc for `omd`. Might also be handy for other purposes. For odoc integration it would at least be cool to have markdown files be converted to ocamldoc and be included in the generated documentation.

nojb:
Got it. Yes, getting to compile with 4.04 should not be too hard. Would you like to open a PR with your local patch?

jfrolich:
Yes!

PR #213, opened 2020-07-03
Downgrade OCaml requirements

jfrolich:
Note this is a very hacky solution. I'd wanted to see if I could make it work, looking to cleanup.

Not sure if there is a better way to do `Fun.protect`. But since it's only used for testing it might be good to have some kind of ifdef/cppo in there.

Closes: #210


src/parser.ml
```diff
@@ -1,5 +1,95 @@
...
 
+module Compat : sig
+  val rep: Uchar.t
+  (** Uchar.rep *)
+
+  val add_utf_8_uchar : Buffer.t -> Uchar.t -> unit
+  (** Buffer.add_utf_8_uchar *)
+
+  val find_opt: ('a -> bool) -> 'a list -> 'a option
+  (** List.find_opt *)
```
nojb, line 11: The way these kinds of shims are usually done is to define a module `Compat` with submodules matching those of the stdlib:
```
module Compat = struct
  module Uchar = struct
    include Uchar
    let rep = ...
  end
  module Buffer = struct
    include Buffer
    let add_utf_8_uchar ... = ...
  end
  ...
end
```
and then simply `open Compat` before the affected code. That way there is no need to touch the actual code that makes use of these functions.

src/parser.ml
```diff
@@ -24,6 +114,7 @@ module Sub : sig
 
   val sub: len:int -> t -> t
 end = struct
+
```
nojb, line 117: Spurious diff.

src/parser.ml
```diff
@@ -1,5 +1,95 @@
...
+  let rep: Uchar.t = Obj.magic(0xFFFD)
+  (* let add_utf_8_uchar = Buffer.add_utf_8_uchar *)
+  type buffer =
+   {mutable buffer : bytes;
+    mutable position : int;
+    mutable length : int;
+    initial_buffer : bytes}
+
+  let get_buffer b =
+    let b = (Obj.magic(b): buffer) in
```
nojb, line 22: `Obj.magic` is infamous for being as safe as juggling with chainsaws... so we will have to do something else here. Instead of duplicating part of the `Buffer` internals here, the idea is to simply reimplement `add_utf_8_uchar` below "from the outside" (using `Buffer.add_char`). It should be fairly easy, but shout if you have any questions.

src/parser.ml
```diff
@@ -470,7 +561,7 @@ let entity s =
             loop (succ m) (n * 16 + Char.code c - Char.code '0') (Sub.tail s)
         | Some ';' ->
             if m = 0 then raise Fail;
-            let u = if n = 0 || not (Uchar.is_valid n) then Uchar.rep else Uchar.of_int n in
+            let u = if n = 0 || not (Uchar.is_valid n) then Compat.rep else Uchar.of_int n in
```
nojb, line 564: This change (and the ones below) are not needed if we organize `Compat` as suggested above.

tests/extract_tests.ml
```diff
@@ -26,13 +26,11 @@ let disabled =
 
 let with_open_in fn f =
   let ic = open_in fn in
-  Fun.protect ~finally:(fun () -> close_in_noerr ic)
-    (fun () -> f ic)
+    f ic
```
nojb, line 29: Since this is just for the test harness, we can do away with the `protect` call, but you should at least call `close_in ic` (and `close_out oc` below) in order to keep up appearances :)

nojb: Also, the indentation seems a bit off.

nojb:
> Not sure if there is a better way to do `Fun.protect`. But since it's only used for testing it might be good to have some kind of ifdef/cppo in there.

Since it is just for the test harness it is not very important. But in general, preprocessing is evil :) so unless we had a very good reason, we should try to avoid having it.

src/parser.ml
jfrolich: Ah makes sense! Will do!

src/compat.ml
```diff
@@ -0,0 +1,32 @@
+module Uchar = struct
+  include Uchar
+  let rep: Uchar.t = Obj.magic(0xFFFD)
```
nojb, line 3: Let's not use `Obj.magic`, rather you can use `Uchar.of_int`.

tests/omd.ml
```diff
@@ -7,8 +7,9 @@ let normalize_html s =
 
 let with_open_in fn f =
   let ic = open_in fn in
-  Fun.protect ~finally:(fun () -> close_in_noerr ic)
-    (fun () -> f ic)
+  let result = f ic in
+  close_in_noerr ic;
+  result
```
nojb, line 12: As this is used in the main library code, let's be a bit more careful, and use the following definition for `protect`:
```
let protect ~finally f =
  match f () with
  | exception e -> finally (); raise e
  | r -> finally (); r
```

jfrolich: Yes I think it should be protected in both cases. Btw I tried to create a compact module to share between `extract_tests` and `omd` and got this error.
```
Error: Module "Compat" is used in several stanzas:
- tests/dune:1
- tests/dune:13
To fix this error, you must specify an explicit "modules" field in every
library, executable, and executables stanzas in this dune file. Note that
each module cannot appear in more than one "modules" field - it must belong
to a single library or executable.
```

I think it's not too bad to have this short function in both files as an alternative.

nojb: Yes, it OK to duplicate the definition.

tests/extract_tests.ml
```diff
@@ -26,12 +30,15 @@ let disabled =
 
 let with_open_in fn f =
   let ic = open_in fn in
-  Fun.protect ~finally:(fun () -> close_in_noerr ic)
+  protect ~finally:(fun () -> close_in_noerr ic)
     (fun () -> f ic)
 
+
+
+
```
nojb, line 38: Spurious diff.

tests/omd.ml
```diff
@@ -7,9 +11,10 @@ let normalize_html s =
 
 let with_open_in fn f =
   let ic = open_in fn in
-  Fun.protect ~finally:(fun () -> close_in_noerr ic)
+  protect ~finally:(fun () -> close_in_noerr ic)
     (fun () -> f ic)
 
+
```
nojb, line 17: Spurious diff.

src/compat.ml
```diff
@@ -0,0 +1,32 @@
...
+  | u when u <= 0xFFFF ->
+      Buffer.add_char b (Char.unsafe_chr (0xE0 lor (u lsr 12)));
+      Buffer.add_char b (Char.unsafe_chr (0x80 lor ((u lsr 6) land 0x3F)));
+      Buffer.add_char b (Char.unsafe_chr (0x80 lor (u land 0x3F)))
+  | u when u <= 0x10FFFF ->
+      Buffer.add_char b (Char.unsafe_chr (0xF0 lor (u lsr 18)));
+      Buffer.add_char b (Char.unsafe_chr (0x80 lor ((u lsr 12) land 0x3F)));
+      Buffer.add_char b (Char.unsafe_chr (0x80 lor ((u lsr 6) land 0x3F)));
+      Buffer.add_char b (Char.unsafe_chr (0x80 lor (u land 0x3F)))
+    | _ -> assert false
```
nojb, line 31: Misindented line.

nojb:
Also, please update the `README` which currently says that 4.09 is required. Thanks!

src/compat.ml
jfrolich: I am not used to write code without a formatted anymore 🤷‍♂️

nojb: Yes, I have been thinking switching to ocamlformat, but let us leave that for a different PR :)

jfrolich:
Ok thanks for your great comments. This is literally the first time I am writing pure OCaml syntax (I am familiar with the AST and Reason syntax).

.github/workflows/workflow.yml
```diff
@@ -14,6 +14,7 @@ jobs:
           - ubuntu-latest
           - windows-latest
         ocaml-version:
+          - 4.04.2
```
nojb, line 17: It would be nice to test all intermediate versions `4.05`, `4.06`, `4.07` as well, but we are going to end up with a rather large test matrix. We could remove, say, `macos-latest` from the test matrix. What do you think?

nojb:
Also, please add an entry to the `CHANGES.md` file with the following format:
```
- Lower OCaml requirement to 4.04.2  (#213, @jfrolich)
```

.github/workflows/workflow.yml
jfrolich: I think it would be good to not remove macOS. As the builds run in parallel it's not a big problem I think (and it's free :))

nojb:
All checks passed. Do you have anything else left for this PR? Otherwise, we can merge it.

CHANGES.md
```diff
@@ -4,6 +4,8 @@ next (2.0.0)
 - Big refactoring by @nojb. Changes in interface (simplification of code, might
   affect performance a little). To be tested!
 
+- Lower OCaml requirement to 4.04.2  (#213, @jfrolich)
+
```
nojb, line 8: Actually, this entry should go above the previous entry, so that we have most-recent-first. Thanks!

nojb:
Thanks! LGTM. Are you happy with the PR? If yes, I will merge.

jfrolich:
Yep!

nojb:
Merged. Thank you!

Issue #216, opened 2020-07-05
The AST representaiton of images and links seems inaccurate


shonfeder:
If not inaccurate (perhaps there is no objective standard in matters of AST representations), it is at least counterintuitive (to me) and awkward to work with.

The markdown spec we test against says that

```
![foo](/url "title")
```

Should be parsed into

```
<p><img src="/url" alt="foo" title="title" /></p>
```

So, the content of the square braces should be the value of the `alt` attribute in the image tag. This seems to hold without any exception.

However, the current representation in the `Omd` ast is:

```
# Omd.of_string "![foo](/url \"title\")";;
- : Omd.doc =
[{Omd.bl_desc =
   Omd.Paragraph
    {Omd.il_desc =
      Omd.Image
       {Omd.label = {Omd.il_desc = Omd.Text "foo"; il_attributes = []};
        destination = "/url"; title = Some "title"};
     il_attributes = []};
  bl_attributes = []}]
```

The value of the `label` field in the `image` record is `Omd.inline`. But that means valid values for this could include raw html, links, soft breaks, etc., when, judging from the spec, it looks like this should always just be a string.E.g., in this example:

```
![foo [bar](/url)](/url2)
.
<p><img src="/url2" alt="foo bar" /></p>
```

Moreover, as far as I can tell, it is impossible for the `li_attributes` field to ever have any use in this context. 

If the label is omitted, we get an even more confusing representation, in which the absence of a label is represented in the presence of an inline concatenation of no elements:

```
# Omd.of_string "![](/url \"title\")";;
- : Omd.doc =
[{Omd.bl_desc =
   Omd.Paragraph
    {Omd.il_desc =
      Omd.Image
       {Omd.label = {Omd.il_desc = Omd.Concat []; il_attributes = []};
        destination = "/url"; title = Some "title"};
     il_attributes = []};
  bl_attributes = []}]
```

IMO, from the perspective of the AST as a UI tool this inaccurate representation isn't worth whatever might be gained by sharing the `link_def` record. A more user friendly representation would be something like: 

```
type image = {alt: string option; title: string option; destination: string}
```

If you're open to this change, I'm happen to open a PR for it.

shonfeder:
I'm just noticing some other surprising behavior, which also strikes me as very inconvenient, and not accurate to the semantics of the data model underlying markdown, based on my understanding. (Of course, it's possible this is because you intend something for this library which is different than I expect, or that I misunderstand in some other way!)

[...]

PR #217, opened 2020-07-05
Get rid of Link_def node

nojb:
As [observed](https://github.com/ocaml/omd/issues/216#issuecomment-653841374) by @shonfeder, the `Link_def` node is no longer useful in the AST. Let's remove it.

### DIFF
--- a/src/ast.ml
+++ b/src/ast.ml
@@ -1,13 +1,6 @@
 type attributes =
   (string * string) list
 
-type 'a link_def =
-  {
-    label: 'a;
-    destination: string;
-    title: string option;
-  }
-
 type list_type =
   | Ordered of int * char
   | Bullet of char
@@ -21,6 +14,14 @@ let same_block_list_kind k1 k2 =
   | Ordered (_, c1), Ordered (_, c2)
   | Bullet c1, Bullet c2 -> c1 = c2
   | _ -> false
+
+type link_def =
+  {
+    label: string;
+    destination: string;
+    title: string option;
+    attributes: attributes;
+  }
 
 module type T = sig
   type t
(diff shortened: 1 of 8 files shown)


jfrolich:
Why is it no longer useful? Isn't it good to have the link definitions in the AST to disambiguate between the two uses of links? Especially if we have an AST -> markdown printer (you probably want to preserve the defs in that case).

nojb:
> Why is it no longer useful? Isn't it good to have the link definitions in the AST to disambiguate between the two uses of links? 

It is no longer useful because reference links are resolved at parsing time since #209. Since we do not keep a reference  to the link definition in the AST, having the link definition node does not serve any purpose.

There is the larger question of how lossy/lossless do we want the AST to be. Lately I have been leaning towards making the AST easier to use (and so more lossy). Preserving link reference and definition nodes allows to reconstruct the original markdown, but makes the AST harder to use, as the user needs to resolve link references.

> Especially if we have an AST -> markdown printer (you probably want to preserve the defs in that case).

Indeed, if we add such a printer, it won't be able to recreate link references, and it will just inline the link. I am not too worried about this (but opinions welcome). The only use I can think of for the AST->markdown printer (apart from pretty printing) is for generating markdown programmatically; the use of link references/definitions does not seem to be crucial for this.

nojb:
As you observed, another case of lossy-ness in the AST is that we do not record which delimiter was used for emphasis/strong emphasis (`_` or `*`) or the number of backquotes used to delimit code blocks (this used to be the case, but added noise to the AST).

nojb:
#223

jfrolich:
I understand you want to reduce complexity, but the case can be made that link definitions are semantically different (so can be treated differently in a particular printer). About the star or underscore emphasis a case can be made that there is no difference, but it can be pretty handy for a printer to allow supporting a richer feature set while not extending the syntax. These add a bit of complexity to the AST, but it can also easily be ignored by printers? 

An addition that I don't really understand is the attributes that are added to the AST.

nojb:
> An addition that I don't really understand is the attributes that are added to the AST.

This is an extension of the standard used by Pandoc which is quite handy to enrich the AST with custom metadata. In the HTML printer, they correspond to usual attributes. See https://github.com/jgm/commonmark-hs/blob/master/commonmark-extensions/test/attributes.md for the spec.

jfrolich:
Ah ok thanks for referencing, I was wondering where that was coming from. That can definitely be handy indeed.

nojb:
Thanks for the feedback re: keeping link references in the AST. I want to mull it over, but I see that it could be useful in some cases.

PR #218, opened 2020-07-05
Downgrade dune requirements to dune 2.0

### LINKED ISSUE #212: Lower dune requirement?
I noticed that the master branch of omd has ` "dune" {>= "2.5"}`

Is there a reason why this is needed?

It makes it more difficult for some packages or internal systems that have to be pinned to an earlier version of Dune, just as one example. Is there something really advanced going on in the build that requires >=2.5?

nojb:
Fixes #212 ?

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -1,4 +1,4 @@
-(lang dune 2.5)
+(lang dune 2.0)
 
 (name omd)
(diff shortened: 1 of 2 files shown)


shonfeder:
Looks like this one is ready to merge!

PR #219, opened 2020-07-05
Esy build sandboxes

jfrolich:
I noticed the builds are taking very long. With esy sandboxes it's possible to cache the build artifacts and make very fasts builds. This will also help with a larger matrix as the total build time will not be that long.


nojb:
Thanks! To be honest I am not too unhappy with the CI times currently. Sometimes it takes longer, but generally it is done within ~15m which seems fine to me. Also, 4.4 is not being tested, which sort of defeats the purpose of #213 :)

I also feel it is important to continue to test the usual OPAM build as well. Given all this, I would rather not switch to `esy` for CI for the time being. Thanks anyway!

jfrolich:
Builds can be reduced to ~2 minutes. I am looking how to support 4.4 and 4.5.

jfrolich:
Just seeing if I can make this work in an elegant manner. There shouldn't be any difference compared to an opam build.

nojb:
> Builds can be reduced to ~2 minutes. I am looking how to support 4.4 and 4.5.

OK, that would be nice! Another alternative is to have both workflows in parallel, with the non-`esy` workflow reserved just for one or two versions/systems that we care in particular.

jordwalke:
For Reason we run our CI on both esy and opam just to make sure everything will be as compatible as possible. Seems like a good idea for a heavily used library like omd.

jfrolich:
Because of the limited availability of ocaml versions on easy and since we need opam anyway let's close this.

nojb:
Sure, makes sense.

PR #222, opened 2020-07-06
Opam cache

jfrolich:


jfrolich:
I managed to get build time for macOS and Linux down to about 1 minute with caching. Still trying to find a way to do the same with windows (but there might be an inherent issue with windows, see: https://github.com/avsm/setup-ocaml/issues/1).

.github/workflows/workflow.yml
```diff
@@ -14,20 +17,51 @@ jobs:
...
+      - name: Get opam root path
+        id: opam
+        run: echo "$(opam var root)"
+
+      - name: Try to restore build cache
+        id: build-cache
+        uses: actions/cache@v2
+        with:
+          path: "_build"
+          key: ${{ matrix.os }}-${{ matrix.ocaml-version }}
```
nojb, line 63: What does the "build cache" cache?

jfrolich: It caches the build artifacts. Isn't really necessary I think!

nojb: Indeed, let's remove this. It is not really necessary (the library builds very quickly anyway), it complicates matters, and the less state to carry around the better.

jfrolich: Agree. It's not good to carry that state anyway.

jfrolich:
Somehow I am getting `Cache not found for input keys: windows-cache-4.04.2` even though the cache was created successfully in a previous build on windows. So the cache is not working for windows. I don't have enough time to dive deeper, but this PR is already a huge win in build time since we have a larger matrix since we support more OCaml versions now. So how about getting this merged and try to fix windows caching later?

nojb:
> Somehow I am getting `Cache not found for input keys: windows-cache-4.04.2` even though the cache was created successfully in a previous build on windows. So the cache is not working for windows. I don't have enough time to dive deeper, but this PR is already a huge win in build time since we have a larger matrix since we support more OCaml versions now. So how about getting this merged and try to fix windows caching later?

Sounds good. Let me know when you think the PR is ready for a review. Thanks!

jfrolich:
I am trying a few things out. Once I'm done I will squash and force push so we have better history.

nojb:
> I am trying a few things out. Once I'm done I will squash and force push so we have better history.

Sure. I can squash everything together when merging, so don't worry about the rebase.

jfrolich:
Ok sounds good!

smorimoto:
The cache can't be found because it's probably hitting the cache limit of the repository. After all, caching with dependencies is not realistic at this time. Rather than do that, I plan to create a minimal switch cache (without dependencies) on the setup-ocaml side.

jfrolich:
> The cache can't be found because it's probably hitting the cache limit of the repository.
Are you saying this doesn't work. I think it did work when I tested.

smorimoto:
Rather than not working, you may hit a limit and not save it. By the way, I've seen it sometimes complicate things because it's hard to really manage the cache correctly unless you lock all versions of the opam package at this point or create an opam lock file.

smorimoto:
Note that there is a 5 GB cache size limit. In large projects, it's often difficult to store caches with all 3 platforms. https://github.com/actions/cache#cache-limits

shonfeder:
Hi, @jfrolich ! Just checking in on this one, as I try to get a handle on our PR backlog. Is this in a good place from your perspective, such that I can/should experiment to ensure it's all working as expected and then merge in -- if possible?

shonfeder:
Thanks, btw, for contributing to the maintainability and development time here. It's much appreciated :) :pray:

jfrolich:
I am not really sure. I think there might be a better github action now (this was quite some time ago).

smorimoto:
See https://github.com/avsm/setup-ocaml/pull/66

shonfeder:
Ok! We'll wait until v2 is out for the time being. Thanks!

PR #225, opened 2020-12-13
improve quoting

madroach:


shonfeder:
@madroach if this is still something you thing we should incorporate, would you be willing to

1. Update the PR description with some context and explanation of the need and motivate for this change
2. Fix the tests on this branch (should be a matter of just running `dune test`, ensuring that the changes are what you expect, and then running `dune promote` and committing those changes.

Thank you for your contribution!

PR #229, opened 2021-03-09
dune-project: add lower bound on ocaml

mseri:
The code uses String.split_on_char which was introduced in 4.04.
Should I also regenerate the opam file?


nojb:
Thanks! I belive the `omd.opam` file (which is auto-generated when you do `dune build`) also needs to be updated.

nojb:
> Should I also regenerate the opam file?

Yes, please :)

mseri:
Done :)

nojb:
Thanks!

Issue #232, opened 2021-04-08
omd2: provide table of contents generation


avsm:
(noting a missing feature from the omd 1.x branch, that was used in opam2web)

omd 1.x provided a [Omd.toc](https://github.com/ocaml/omd/blob/1.2.5/src/omd.mli#L151) function for table-of-contents generation. The [implementation](https://github.com/ocaml/omd/blob/1.2.5/src/omd.ml#L121) may be re-usable, with the new AST types.

shonfeder:
Thanks for reporting the omission! I'm considering recovery of this functionality higher priority, but I'd like to settle #223 and #228 first, to minimize rework.

sonologico:
Started working on this on #240.

avsm:
Closed by #240 -- it's certainly enough for ocaml.org to generate its ToC. Thanks @sonologico !

PR #241, opened 2021-05-29
Add ocamlformat and update CI pipeline

shonfeder:
- Add an `.ocamlformat` config
- Reformat code base
- CI pipeline updates
  - Use github actions v2
  - Run formatting check
  - Add missing compiler versions
  - Enable dune-cache

I'll wait to merge this until #232 (and other substantial outstanding PRs?) are merged in.

In the meantime, @sonologico, whenever you have a chance, let me know if you are OK with these formatting, or if you have strong feelings by a different config.

### DIFF
--- a/bin/main.ml
+++ b/bin/main.ml
@@ -1,14 +1,22 @@
 let with_open_in fn f =
   let ic = open_in fn in
   match f ic with
-  | r -> close_in ic; r
-  | exception e -> close_in_noerr ic; raise e
+  | r ->
+      close_in ic;
+      r
+  | exception e ->
+      close_in_noerr ic;
+      raise e
 
 let with_open_out fn f =
   let oc = open_out fn in
   match f oc with
-  | r -> close_out oc; r
-  | exception e -> close_out_noerr oc; raise e
+  | r ->
+      close_out oc;
+      r
+  | exception e ->
+      close_out_noerr oc;
+      raise e
 
 let process ic oc =
   let md = Omd.of_channel ic in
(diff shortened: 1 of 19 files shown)


bin/main.ml
```diff
@@ -1,14 +1,22 @@
 let with_open_in fn f =
   let ic = open_in fn in
   match f ic with
-  | r -> close_in ic; r
-  | exception e -> close_in_noerr ic; raise e
+  | r           ->
+      close_in ic;
+      r
+  | exception e ->
```
sonologico, line 7: Everything looks good to me, except for the aligned arrows. I don't feel strongly about it, so if you really prefer it that way, I'm ok with it.

Having said that, the rule is either not applied consistently or I don't understand when it should apply.

shonfeder: I am ok either way. I agree it's awkward in these cases, but I tend to feel like it pays off when you have lots of similarly sized matches.

I'm not sure what's up with the inconsistency!

I've removed that rule. Let me know if you like this better :)

sonologico: I prefer it this way, but I'm ok with the other way if we figure out why it's inconsistent.

sonologico (approved):
Looks good to me. Thanks for picking this up.

Don't worry about my PR. I'll sort it out. Running ocamlformat on my changes should make the conflicts manageable.

If there's another open PR that you think we should wait on, let's prioritize that. Then we can get this one out of the way and not have to think about formatting anymore.

Issue #235, opened 2021-04-14
Fix all conformance tests


shonfeder:
#233 is a good start at cleaning up the outstanding conformance tests, but we aim to resolve them all. See https://github.com/ocaml/omd/blob/master/tests/extract_tests.ml#L7 for the disabled tests

shonfeder:
I'm gonna dig into this a bit, fyi @SquidDev, since you've also made some progress on this point :) Let me know if my work on this would conflict with ongoing work you have locally.

SquidDev:
I'm afraid I've not had a proper look at this since.

I started redoing the link parser to match the recommended approach in the common mark spec (our handling of nested images/links is one of the big remaining issues), but didn't make significant enough progress that I think it should stop other people having a look.

shonfeder:
Thanks @SquidDev! If you'd be open to sharing your WIP, you might open a draft PR with your work so far? No pressure to do so however!

I ended up refactoring the specifications failures into smaller issues. iiuc, your ongoing work is aimed at #245? Perhaps also #244. Those definitely seem to the be the most significant bugs in the current parsing.

shonfeder:
@tatchi took out the bulk of these! :tada:

PR #247, opened 2021-05-29
Fix spec 353: ASCII character 160, non-breaking space, is white space

shonfeder:
Fixes failing verification of spec unit 353. See 2441975 for details.

Turned up while refactoring issue #235 into smaller work items.

This PR also adds make targets to streamline regeneration of tests and


sonologico (approved):
Looks good. Good catch.

Issue #221, opened 2020-07-05
Definition lists do not implement pandoc spec fully


nojb:
Pandoc spec:
https://github.com/jgm/commonmark-hs/blob/master/commonmark-extensions/test/definition_lists.md

- Loose vs tight lists not taken into account
- Multi block definitions not supported
- `~` cannot be used in place of `:`

cc @clecat

clecat:
Thank you for your feedback, I might be a bit busy this week, but I'll try to do it asap. I suppose the full implementation of the spec is mandatory for this feature to be in the release

nojb:
> Thank you for your feedback, I might be a bit busy this week, but I'll try to do it asap. I suppose the full implementation of the spec is mandatory for this feature to be in the release

The issue was just to document what is missing, it is not urgent. As for the release, we could just delay definition lists until after the release, it is not a problem (I think it is better to have a complete implementation before it is released).

nojb:
No plans to address this, let's close.

PR #257, opened 2021-10-30
Make definitions block-level

talex5:
I just converted my blog from Octopress to omd, and I needed support for definition lists with multiple paragraphs.

Not completely sure what I'm doing here, but this changed got it working it for me, and should go some way towards #221.

### DIFF
--- a/src/block.ml
+++ b/src/block.ml
@@ -21,7 +21,12 @@ module Pre = struct
         * attributes
     | Rindented_code of string list
     | Rhtml of Parser.html_kind * string list
-    | Rdef_list of string * string list
+    | Rdef_list of
+        string
+        * int
+        * bool  (* true if we've seen a blank line - disables lazy mode) *)
+        * attributes Raw.block list list
+        * t
     | Rempty
 
   and t =
@@ -69,13 +74,25 @@ module Pre = struct
         Code_block (attr, label, "") :: blocks
     | Rfenced_code (_, _, _kind, (label, _other), l, attr) ->
         Code_block (attr, label, concat l) :: blocks
-    | Rdef_list (term, defs) ->
-        let l, blocks =
-          match blocks with
-          | Definition_list (_, l) :: b -> (l, b)
-          | b -> ([], b)
+    | Rdef_list (term, _, _, defs, state) ->
+        let def = finish state in
+        let defs = def :: defs in
+        let this_sp =
+          match def with
+          | [Paragraph _] -> Tight
+          | _ -> Loose
         in
-        Definition_list ([], l @ [ { term; defs = List.rev defs } ]) :: blocks
+        let l, sp, blocks =
+          match blocks with
+          | Definition_list (_, sp, l) :: b ->
+              let sp = match sp, this_sp with
+                | Tight, Tight -> Tight
+                | Loose, _ | _, Loose -> Loose
+              in
+              (l, sp, b)
+          | b -> ([], this_sp, b)
+        in
+        Definition_list ([], sp, l @ [ { term; defs = List.rev defs } ]) :: blocks
     | Rindented_code l ->
         (* TODO: trim from the right *)
         let rec loop = function
@@ -116,10 +133,10 @@ module Pre = struct
         }
     | Rempty, (Lsetext_heading _ | Lparagraph | Ldef_list _) ->
         { blocks; next = Rparagraph [ Sub.to_string s ] }
-    | Rparagraph [ h ], Ldef_list def ->
-        { blocks; next = Rdef_list (h, [ def ]) }
-    | Rdef_list (term, defs), Ldef_list def ->
-        { blocks; next = Rdef_list (term, def :: defs) }
+    | Rparagraph [ h ], Ldef_list (indent, def) ->
+        { blocks; next = Rdef_list (h, indent, false, [], process empty (Sub.of_string def)) }
+    | Rdef_list (term, _, false, defs, state), Ldef_list (indent, def) ->
+        { blocks; next = Rdef_list (term, indent, false, finish state :: defs, process empty (Sub.of_string def)) }
     | Rparagraph _, Llist_item ((Ordered (1, _) | Bullet _), _, s1)
       when not (Parser.is_empty (Parser.P.of_string (Sub.to_string s1))) ->
         process { blocks = close { blocks; next }; next = Rempty } s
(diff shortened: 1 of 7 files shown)


src/omd.mli
```diff
@@ -30,18 +30,18 @@ and 'attr inline =
...
+and 'attr block =
   | Paragraph of 'attr * 'attr inline
   | List of 'attr * list_type * list_spacing * 'attr block list list
   | Blockquote of 'attr * 'attr block list
   | Thematic_break of 'attr
   | Heading of 'attr * int * 'attr inline
   | Code_block of 'attr * string * string
   | Html_block of 'attr * string
-  | Definition_list of 'attr * 'attr def_elt list
+  | Definition_list of 'attr * list_spacing * 'attr def_elt list
```
shonfeder, line 44: If this is to become a recursive type, perhaps the `block` should be listed first, since it's the "principle" type?

tests/def_list.md
```diff
@@ -10,13 +10,80 @@ Second Term
...
+# With multiple paragraphs
+
+First Term
+: This is the definition of the first term.
+
+Second Term
+: This is one paragraph of the second term.
+
+  This is another paragraph of the second term.
+
```
shonfeder, line 23: Perhaps we should include cases for

- mixing lazy multiline and the multi-paragraph markup
- mixing using multiple definition terms with the multi-paragraph markup
- some case like

   ```
   # foo
   : not a definition
   ```

shonfeder (commented):
Thanks for the improvement! This will definitely be useful.

I've a few suggestion, but I'd be happy to make them on this PR if you're not inclined to. Just let me know if you'd prefer that.

shonfeder:
Looks like we'll also need to run ocamlformat, btw.

talex5:
> I've a few suggestion, but I'd be happy to make them on this PR if you're not inclined to. Just let me know if you'd prefer that.

That would be great - thanks! Your suggestions all sound sensible to me.

tests/def_list.md
```diff
@@ -10,13 +10,122 @@ Second Term
...
+<p>This is one paragraph using lazy wrapping,
+so this should be part of the first paragraph
+of the third term.</p>
+<p>But this should be the second paragraph of
+the second paragraph</p>
+<p>Note that, currently, subsequent paragraphs cannot lazy wrap.</p>
+</dd>
+</dl>
+<p>For example, this line will be part of a new paragraph outside
+of the definition list entirely.</p>
```
shonfeder, line 96: @talex5 does this seem like the right behavior to you? I'm not exactly sure what ought to happen in this case. I can report that in pandoc, this markdown

```
Term 1
: A first paragraph that is
lazily wrapped

    A second paragrap that 
is lazily wrapped 
```

is rendered as 

```html
<dl>
<dt>Term 1</dt>
<dd><p>A first paragraph that is lazily wrapped</p>
<p>A second paragrap that is lazily wrapped</p>
</dd>
</dl>
```

But they use 4 spaces to indicate the start of the second paragraph.

shonfeder (commented):
Pushed my changes up. I do have one pending question for you about some edge case behavior after I added a few more things to the test cases.

src/block.ml
```diff
@@ -147,10 +181,34 @@ module Pre = struct
...
+            Rdef_list
+              { deflist with
+                empty_line_seen = true
+              ; state = process deflist.state s
+              }
         }
+    | Rdef_list deflist, _ when Parser.indent s >= deflist.indent ->
+        let s = Sub.offset deflist.indent s in
+        let state = process deflist.state s in
+        { blocks; next = Rdef_list { deflist with state } }
```
talex5, line 207: I guess this should set `empty_line_seen = false` to fix the problem with lazy wrapping you mention below.

src/block.ml
```diff
@@ -147,10 +181,34 @@ module Pre = struct
         { blocks
         ; next = Rfenced_code (ind, num, q, info, Sub.to_string s :: lines, a)
         }
-    | Rdef_list (term, d :: defs), Lparagraph ->
+    | ( Rdef_list ({ empty_line_seen = false; _ } as deflist)
```
talex5, line 184: Thinking about this a bit more, I don't think we need the `empty_line_seen = false` check here.

src/block.ml
shonfeder: This does address the lazy wrapping, thanks for the tip! Unfortunately it also breaks the nested def-list parsing. I'll continue considering it, but if an immediate solution occurs to you, let me know ;)

src/block.ml
shonfeder: Changing this (either by removing the match on the `empty_line_seen` or removing the case all together) leads to failures around the `This is not a correct definition list` line. This to will need some considering. The moral of the story seems to me that the handwritten parsing logic here is quite touchy and tough to reason about...

talex5: https://github.com/jgm/commonmark-hs/blob/master/commonmark-extensions/test/definition_lists.md ("Multiple definitions, loose") gives this example:

```
apple

:   red fruit

:   computer company

orange

:   orange fruit
:   telecom company
.
<dl>
<dt>apple</dt>
<dd>
<p>red fruit</p>
</dd>
<dd>
<p>computer company</p>
</dd>
<dt>orange</dt>
<dd>
<p>orange fruit</p>
</dd>
<dd>
<p>telecom company</p>
</dd>
</dl>
```
So perhaps the `This is not a correct definition list` test is itself wrong?

shonfeder: Ah, thanks for the reference! I'm happy with deferring to commonmark on these things. I'll spend some time comparing their test suite with ours later this week or over the weekend.

shonfeder: Ah, duh. Now that I look at the test case, since you've introduced the loose spacing, of course that test is now incorrect! My bad for not giving it sufficient thought before.

shonfeder: So, if we remove this check, then it fixes the loose format for multiple definitions, but it breaks the nesting. Which I think means that, even if we were OK with the leaving the lazy line wrapping broken for multi-paragraph definitions, we're still a bit stuck on this point.

I'll investigate over the weekend, and try to get my head around properly around the parsing function. But just an FYI to explains why I'm not merging this straight away.

src/block.ml
sonologico: Reordering the clauses such that the one that checks for the indentation level comes first solves the problem with the nested def-list test. The only remaining problem on def_list-003 test then are missing `<p>`s in the generated html.

PR #258, opened 2021-11-06
Preserve order of input file list

patricoferris:
I've been working on something that uses the `omd` binary to provide a text-only version of the markdown as a standalone html page and noticed `omd x y z` outputs HTML to `stdout` in the order `z y x`, I tried `omd.1.3.1` and it outputs the HTML in the same order the files were given which I think is the more expected behaviour.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+unreleased
+----------
+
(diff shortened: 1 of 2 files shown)


shonfeder:
Makes sense! Thanks for the fix! :)

shonfeder (approved):

PR #261, opened 2022-02-20
Fix spec 519. Reprocess inlines in case a label turns out to not be for a link

sonologico:
I'll double check if my intuition is correct, but this fixes the test.

What was happening is that we would process the inline content of the (would be) label and use the results even if it didn't turn out to be a label. However, because one `*` was between brackets and the other `*` was outside, they were never matched and therefore treated as text.


shonfeder (approved):
LGTM!

Please merge when you are satisfied with your solution :)

PR #262, opened 2022-02-26
Spec 334: Preserve spaces in inline code blocks consisting of spaces only

### LINKED ISSUE #243: failing verificaiton of spec unit 334: don't strip spaces when code span only contains spaces
Failing example:

```
` `
`  `
.
<p><code> </code>
<code>  </code></p>
```

for requirement

> No stripping occurs if the code span contains only spaces:

https://github.com/madroach/omd/blob/master/tests/spec.txt#L5923-L5931

Required for #235

sonologico:
Fixes #243

### DIFF
--- a/src/compat.ml
+++ b/src/compat.ml
@@ -51,8 +51,12 @@ module String = struct
   let for_all p s =
     let n = length s in
     let rec loop i =
-      if i = n then true
-      else if p (unsafe_get s i) then loop (succ i)
-      else false in
+      if i = n then
+        true
+      else if p (unsafe_get s i) then
+        loop (succ i)
+      else
+        false
+    in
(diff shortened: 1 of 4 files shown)


src/compat.ml
```diff
@@ -44,3 +44,15 @@ module Buffer = struct
         Buffer.add_char b (Char.unsafe_chr (0x80 lor (u land 0x3F)))
     | _ -> assert false
 end
+
+module String = struct
+  include String
+
+  let for_all p s =
```
sonologico, line 51: `String.for_all` is available from `4.13.0`. This is copied from the stdlib.

shonfeder (approved):
LGTM! :tada:

Issue #231, opened 2021-03-15
Support printing to markdown


shonfeder:
This would allow programmatic construction and transformation of markdown, and (in the special case of a no-op transformation) would serve as a markdown pretty printer.

Request by @jfrolich on at https://github.com/ocaml/omd/issues/230#issuecomment-799033420

cuihtlauac:
Not sure if it is already known. ocaml-lsp-server is stuck with 1.3.2:
```
(omd (and (>= 1.3.2) (< 2.0.0~alpha1)))
```
It's using `Omd.to_markdown` in doc_to_md.ml

I believe this is problematic.

shonfeder:
@cuihtlauac Could you explain what kind of problems you're hitting on 1.3.2?

cuihtlauac:
Because `ocaml-lsp-server` requires `omd < 2` it is not possible to have it installed in the same switch as packages requiring `omd >= 2`. Being forced to choose between mutually exclusive dependencies is a problem, I believe.

shonfeder:
Thanks for explaining. I think this issue is certainly top priority at the moment.

My suggestion for the moment would be to avoid depending on the pre-release versions of 2.0 for any production-grade projects. As noted in the README, we have not yet achieved feature parity with 1.3, and only just recently resolved the lingering bugs causing failures to satisfy the CommonMark spec.

cuihtlauac:
You're right. Unfortunately, a number of projects have decided not to wait and are relying on 2.0 features, they probably had good reasons too.

Issue #223, opened 2020-07-06
Add a parse tree structure (a CST) in addition to the AST in oder to preserve all relevant details of the original markdown


nojb:
Should the AST preserve as much as possible of the original markdown source, making it a bit more inconvenient to use, or should it strive for convenience and try to simplify as much as possible of the original document?

Features that are currently simplified in the AST (there may be a couple more, will amend if necessary):

- links are inlined, so references (eg `[foo][bar]`) are not kept in the AST.
- the kind of delimiter used in emphasis/strong emphasis (`_` or `*`) is not kept in the AST.
- the number of backquotes used to delimit a code block is not kept in the AST.

What do you think?

shonfeder:
What about providing two different representations (as suggested in #216)? At the point that we are encoding the location of link references and noting which character is used for emphasis, or how many characters are used to indicate code blocks, I really think it's reasonable to start seeing the representation as capturing the CST. So we could provide a `Cst` module that reflects this kind of stuff, and an `Ast` module (derived by mapping out of the former) that provides a much sleeker interface.

shonfeder:
While study the `spec.txt` I just encountered this specification, which i think helps put a point on this problematic:

> A [link reference definition] does not correspond to a structural element of a document.

The question here, I think, is whether the AST is meant to represent the structure of a document or the structure of markdown syntax. From the perspective of document modeling, thinks like link labels are just accidental complexity introduced for the sake of a particular concrete syntax. But from the perspective of a faithful representation of the flavour of markdown we're supporting, the difference between a reference link and an inline link may be significant.

As a user, I had initially been approaching Omd as something like (a limited) pandoc: it reads markdown into an AST representing a document structure. However, iiuc the direction the package has subsequently evolved is more towards a high-fidelity representation of concrete markdown syntax. I suspect that different users will want these different interfaces, but I don't have any empirical evidence to support that yet (except, perhaps, the success of pandoc. But of course that project has a much wider scope than Omd).

jfrolich:
I think capturing this in the AST would be good because then it's possible to build a markdown printer on the same AST (or converting markdown to something similar). The added complexity wouldn't be too large (I think?).

shonfeder:
@jfrolich that seems like a good motivating example! Just to clarify: I'm pretty convinced we **do** need to preserve this kind of detail, my only qualification here is the consideration of whether we might not *also* want to (eventually) target a more abstract interface.

[...]

PR #265, opened 2022-03-10
Add markdown printer

yasunariw:
See #231.

This adds a printer that transforms the AST back into Markdown text.

Please let me know if there is a good way of adding tests to the current setup.


patricoferris:
Cool feature @yasunariw ! 

I worked on something similar and you can see my findings here: https://github.com/ocaml/omd/issues/223#issuecomment-893821434. I think the conversion from `string -> Omd.doc` losses information so it is quite tricky to implement a `to_markdown` function. For example, there are some corner cases such as:

```ocaml
utop # Omd.of_string "_*hello*_" |> Omd.to_markdown;;
- : string = "**hello**\n"
```

Note this is after pinning `e60a510a54da4c3ed63da34fe6a35f6287570989`. Here, it is valid markdown to nest emphasis using the two different characters, but after using the `Omd.to_markdown` function it has been transformed into `strong` text.

What might be useful is the (very hacked together) testsuite I was running here: https://github.com/patricoferris/omd/tree/omd-print/tests. The idea was to parse the markdown normally and get the HTML, then use the `to_markdown` function and parse that output to HTML and compare the two HTML outputs. It is probably not full-proof but it was how I started finding these edge cases.

yasunariw:
Thanks for all the pointers @patricoferris ! I'll see what else I catch using your test method and report back.

shonfeder:
Thanks for the work on this important feature, @yasunariw and @patricoferris.

> Please let me know if there is a robust way to add tests to the current setup. Thanks!

The most robust way I can think of is via some property based testing. If we had generator that constructed valid arbitrary `Omd.doc` values, then we could check that forall `x : Omd.doc`, `(x |> Omd.to_markdown |> Omd.of_string) =  x`. This should catch the kind of edge case that @patricoferris has pointed out, iiuc.

My inclination is to suggest that we should expect any mergable solution to this feature to dependent on #223.

src/markdown.ml
```diff
@@ -0,0 +1,341 @@
...
+  let n = infer_num_backticks 3 c in
+  nchar n buf '`';
+  let label = String.trim label in
+  Buffer.add_string buf label;
+  add_attrs_to_buffer ~space:true buf attr;
+  nl buf;
+  add_prefix ~first_item_prefix ~prefix buf;
+  add_string_with_nl_prefix ~first_item_prefix ~prefix buf c;
+  let n = String.length c in
+  if n > 0 && c.[String.length c - 1] <> '\n' then begin
```
samoht, line 290:
```suggestion
  let n' = String.length c in
  if n' > 0 && c.[n'- 1] <> '\n' then begin
```

samoht: otherwise the `nchar n buf` bellow adds too much backticks when closing the code block.

Issue #251, opened 2021-08-30
(Option to) Automatically fill the `id` attributes of the headings


smondet:
It's implemented in `ood`:

- `slugify`: https://github.com/ocaml/ood/blob/7b2e15544ba90cca5f6a5baf606a3761ed62742b/src/ood-gen/lib/utils.ml#L37 (can be easily reimplemented without `str`)
- `doc_with_ids`: https://github.com/ocaml/ood/blob/7b2e15544ba90cca5f6a5baf606a3761ed62742b/src/ood-gen/lib/tutorial.ml#L36-L64

it'd be nice to have in `omd`, also to make sure there is a bit of standardization (maybe choose same IDs as pandoc).

shonfeder:
I think this is a good idea.

tatchi:
Hey 👋

I would be happy to work on that issue if nobody started yet :)

PR #267, opened 2022-05-03
automatically fill the id attributes of the headings

tatchi:
Fixes: #251 

This is my attempt at implementing #251

I didn't add an option to enable/disable that feature. Let me know if I should.


tests/blackbox/heading-id.t
```diff
@@ -0,0 +1,8 @@
+  $ omd << "MD"
+  > ### This is an Header Without Id
+  > ### 1 2 Header that starts with 2 numbers
+  > ### Header with an id {#header-id}
+  > MD
+  <h3 id="this-is-an-header-without-id">This is an Header Without Id</h3>
+  <h3 id="header-that-starts-with-2-numbers">1 2 Header that starts with 2 numbers</h3>
+  <h3 id="header-id">Header with an id</h3>
```
shonfeder, line 8: What's the reason for stripping of leading digits? I'm not an HTML expert, but looking at [the spec for HTML5](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute), it looks like [id attributes can include almost any characters]?:

> There are no other restrictions on what form an ID can take; in particular, IDs can consist of just digits, start with a digit, start with an underscore, consist of just punctuation, etc.

We can probably expand this test to cover some other edge cases, like with some unusual characters (including unicode) mixed in?

shonfeder (approved):
Thanks for the contribution!

I've just one question, but it generally looks solid to me. :tada:

shonfeder (commented):
Meant to leave review status as "Comment" :)

tests/blackbox/heading-id.t
tatchi: Thanks for the review @shonfeder 😊

> What's the reason for stripping of leading digits? I'm not an HTML expert, but looking at [the spec for HTML5](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute), it looks like [id attributes can include almost any characters]?:

Because I noticed that's what pandoc does, and the original issue says: 

> also to make sure there is a bit of standardization (maybe choose same IDs as pandoc).

I'm not familiar with pandoc so please let me know if I misunderstood something, it can be easily changed.

> We can probably expand this test to cover some other edge cases, like with some unusual characters (including unicode) mixed in?

Makes sense, I'll try to come up with a few more test cases.

shonfeder:
Ah, I just looked at the failures in CI. There are two issues:

1. To maintain compatibility with OCaml 4.04.2, we'll either need to avoid using the `Fun` module and write our implementations, or pull in https://github.com/thierry-martinez/stdcompat (or an equivalent, if available). I'm quite happy with doing the latter.
2. The failures on recent version of OCaml show why this actually does have to be configurable: it breaks all of our specification compliance tests :). I suggest we make this configurable by CLI, by default to having it enabled.

How does that sound?

tatchi:
> 1. To maintain compatibility with OCaml 4.04.2, we'll either need to avoid using the `Fun` module and write our implementations, or pull in https://github.com/thierry-martinez/stdcompat (or an equivalent, if available). I'm quite happy with doing the latter.

I just inlined the function as it was straightforward to do: https://github.com/ocaml/omd/pull/267/commits/0570385a3f8da16ed65929c7b4c983803a5e4d68

> 2. The failures on recent version of OCaml show why this actually does have to be configurable: it breaks all of our specification compliance tests :). I suggest we make this configurable by CLI, by default to having it enabled.
> 

I added an `--auto-identifiers` option (inspired by pandoc) in https://github.com/ocaml/omd/pull/267/commits/cb088d1606baf9a9956c7bcd07f2ad8aa1d6bb54. Since it's enabled by default it still break the tests. Should I promote them or disable that option when running the tests?

As for the identifiers, there's actually some [rules](https://pandoc.org/MANUAL.html#extension-auto_identifiers) described in the doc on how to derive them:

> The default algorithm used to derive the identifier from the heading text is:
> 
> * Remove all formatting, links, etc.
> * Remove all footnotes.
> * Remove all non-alphanumeric characters, except underscores, hyphens, and periods.
> * Replace all spaces and newlines with hyphens.
> * Convert all alphabetic characters to lowercase.
> * Remove everything up to the first letter (identifiers may not begin with a number or punctuation mark).
> * If nothing is left after this, use the identifier section.

I had to switch to an implementation that supports Unicode characters (using the `Uucp` and `Uutf` modules). It's not yet perfect as some of the test results are not correct. I'll continue to try to improve that :)

tatchi:
I improved the algorithm slightly; it should be ok for a second review.

There are still some differences compared to the result I get on [https://pandoc.org/try/](https://pandoc.org/try/), but pandoc can only generate id's when using `markdown` and not `commonmark`. So the outputted HTML content (from which the id is derived) might be different in some cases, hence why the result might differ.

The only thing missing I think is:

> several headings have the same text; in this case, the first will get an identifier as described above; the second will get the same identifier with -1 appended; the third with -2; and so on.

So for instance in pandoc, the following

```
### hello
### hello
```
becomes:
```
<h3 id="hello">hello</h3>
<h3 id="hello-1">hello</h3>
```

I guess it means that we'll need to keep a state somewhere with all the ids we've already generated. Not sure how to best implement that and if that's something we want. Would love to have your input on that :) 

We'll also have to decide how to handle "empty" id. In pandoc:

> If nothing is left after this, use the identifier section.

So
```
### 33
### 22
### 
```
becomes

```
<h3 id="section">33</h3>
<h3 id="section-1">22</h3>
<h3 id="section-2"></h3>
```

tmattio:
cc @koonwen, this would probably fix https://github.com/ocaml/ocaml.org/issues/523!

src/html.ml
```diff
@@ -128,9 +178,13 @@ and inline = function
   | Image (attr, { label; destination; title }) ->
       img label destination title attr
 
-let rec block = function
+let rec block ~auto_identifiers = function
```
shonfeder, line 181: > I guess it means that we'll need to keep a state somewhere with all the ids we've already generated. Not sure how to best implement that and if that's something we want. Would love to have your input on that :)

Perhaps we can have the `block` function take a record for this argument that can carry some configuration/state? WDYT?

shonfeder:
> We'll also have to decide how to handle "empty" id. In pandoc:

I'm find with following pandoc here.

src/omd.ml
```diff
@@ -16,7 +16,10 @@ let parse_inlines (md, defs) =
 
 let of_channel ic = parse_inlines (Pre.of_channel ic)
 let of_string s = parse_inlines (Pre.of_string s)
-let to_html doc = Html.to_string (Html.of_doc doc)
+
+let to_html ?auto_identifiers doc =
+  Html.to_string (Html.of_doc ?auto_identifiers doc)
```
shonfeder, line 21: In order to run our tests with this flag set to false, I think we can just set the `auto_identifiers` flag to `false` in the call to `to_html` here:

https://github.com/ocaml/omd/blob/191414e6525a811803dd5285dc07848882859c88/tests/omd.ml#L20-L22

shonfeder (commented):
Looking good! I finally was able to address your outstanding questions. Sorry again for the long delay here!

src/html.ml
tatchi: Thanks for the suggestion, I added an `int StringMap.t` to keep track of the numerical suffix. Let me know what you think :)

src/html.ml
```diff
@@ -80,6 +80,78 @@ let escape_uri s =
...
+    | `Uchar u when is_white_space u && !last_is_ws -> ()
+    | `Uchar u ->
+        (if is_alphabetic u || is_hex_digit u then
+         match Uucp.Case.Map.to_lower u with
+         | `Self -> add_to_buffer u
+         | `Uchars us -> List.iter add_to_buffer us);
+        if u = underscore || u = hyphen || u = period then add_to_buffer u
+  in
+  Uutf.String.fold_utf_8 fold () s;
+  Buffer.contents b
```
shonfeder, line 153: This is another case where we can use the accumulator argument to avoid needing a mutable ref. I've added a suggestion to my branch at https://github.com/ocaml/omd/commit/ba7af24be5c416f581cac96facc988c51ec96866 . IMO, while I prefer my suggestion, it's not as obvious that it is an improvement. So this bit is just a suggestion rather than a change request. You might find something in that commit worth incorporating tho, even if you prefer to stick with the mutable ref.

src/html.ml
```diff
@@ -178,7 +254,36 @@ let rec block = function
       in
       elt Block "dl" attr (Some (concat_map f l))
 
-let of_doc doc = concat_map block doc
+let of_doc ?(auto_identifiers = true) doc =
+  let identifiers = Identifiers.empty in
```
shonfeder, line 258: This way of tracking looks pretty good to me! It occurs to me looking at this now that we may want to actually add these into the AST, rather than only producing them when rending HTML. But that's something we can do as a followup, by moving this into the parser.

src/html.ml
```diff
@@ -178,7 +254,36 @@ let rec block = function
...
+          if (not auto_identifiers) || List.mem_assoc "id" attr then
+            (attr, identifiers)
+          else
+            let id = slugify (to_plain_text text) in
+            (* Default identifier if empty. It matches what pandoc does. *)
+            let id = if id = "" then "section" else id in
+            let count, identifiers = Identifiers.touch id identifiers in
+            let id =
+              if count = 0 then id else Printf.sprintf "%s-%i" id count
+            in
```
shonfeder, line 271: I feel like this might better belong in the `slugify` function. WDYT?

shonfeder (commented):
Just a few followup suggestions. I think we're just about there!

tests/blackbox/heading-id.t
```diff
@@ -0,0 +1,32 @@
...
+  <h3 id="ôhey-1">👋👋 ÔHey! 👋👋</h3>
+  <h3 id="dogs--in-my-house"><em>Dogs</em>?--in <em>my</em> house?</h3>
+  <h3 id="html-s5-or-rtf">[HTML], [S5], or [RTF]?</h3>
+  <h3 id="applications">3. Applications</h3>
+  <h3 id="hello.world">hello.world</h3>
+  <h3 id="hello-">-hello-</h3>
+  <h3 id="with-multiple-spaces">with    multiple   spaces</h3>
+  <h3 id="with-spaces">with   spaces</h3>
+  <h3 id="section">33</h3>
+  <h3 id="section-1"></h3>
```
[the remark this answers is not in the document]
tatchi: Definitely makes sense, thanks!

src/html.ml
tatchi: Do you mean adding the id attributes to the AST? If so, yes I think it would make sense.

src/html.ml
tatchi: I don't have any strong opinion on that one. I made the changes in [`820f310` (#267)](https://github.com/ocaml/omd/pull/267/commits/820f310bc9c7621aab419ba2c091968d1b954423)

src/html.ml
shonfeder: That's right: I mean rather than only generating IDs during output to HTML, we may want to add an option to add generated id attributes during parsing, or as a post-processing stage given a parsed AST.

This would expose the functionality both to other output formats, and to inspection by code consuming the AST.

bin/main.ml
```diff
@@ -33,11 +33,15 @@ let print_version () =
...
 let output = ref ""
+let auto_identifiers = ref None
 
 let spec =
   [ ( "-o"
     , Arg.Set_string output
     , " file.html Specify the output file (default is stdout)." )
+  ; ( "--auto-identifiers"
+    , Arg.Bool (fun x -> auto_identifiers := Some x)
+    , " Should identifiers be automatically assigned to headings." )
```
shonfeder, line 44:
```suggestion
    , "Automatically generate identifiers for headings in HTML output." )
```

shonfeder (approved):
Looks great. Thanks so much, and sorry for the sluggish review loops!

shonfeder:
Closing this in favor of #294. I opened up the new branch purely for the purposes of making merge conflict resolution easier.

Thanks so much, @tatchi!

src/html.ml
tatchi: Sorry for the delay, I created an issue to track it https://github.com/ocaml/omd/issues/296

Issue #252, opened 2021-09-24
Constructors for values of type `Omd.doc`


shonfeder:
It can be a bit tedious to construct subtrees in the AST. I think some constructors along the following lines would be helpful:

```ocaml

val txt : ?attrs:attributes -> string -> attributes inline 
val em : ?attrs:attributes -> string -> attributes inline
(* ... *)
val pg : ?attrs:attributes -> attributes inline -> attributes block
val ol :  ?spacing:list_spacing -> ?attrs:attributes -> attributes list list block -> attributes block
val ul :  ?spacing:list_spacing -> ?attrs:attributes -> attributes list list block -> attributes block
```

PR #268, opened 2022-05-03
add constructors for values of type `Omd.doc`

tatchi:
Fixes: #252

Not sure if all the constructors make sense. Open to suggestions, also regarding the functions' names.

### DIFF
--- a/src/omd.ml
+++ b/src/omd.ml
@@ -14,6 +14,41 @@ let parse_inlines (md, defs) =
   in
   List.map (Mapper.map (parse_inline defs)) md
 
+let txt ?(attrs = []) s = Text (attrs, s)
+
+let em ?(attrs = []) il = Emph (attrs, il)
+
+let strong ?(attrs = []) il = Strong (attrs, il)
+
+let code ?(attrs = []) s = Code (attrs, s)
+
+let hard_break attrs = Hard_break attrs
+
+let soft_break attrs = Soft_break attrs
+
+let link ?(attrs = []) li = Link (attrs, li)
+
+let img ?(attrs = []) li = Image (attrs, li)
+
+let html ?(attrs = []) s = Html (attrs, s)
+
+let pg ?(attrs = []) il = Paragraph (attrs, il)
+
+let ul ?(attrs = []) ?(spacing = Loose) l = List (attrs, Bullet '-', spacing, l)
+
+let ol ?(attrs = []) ?(spacing = Loose) l =
+  List (attrs, Ordered (1, '.'), spacing, l)
+
+let blq ?(attrs = []) blocks = Blockquote (attrs, blocks)
+
+let hr attrs = Thematic_break attrs
+
+let code_block ?(attrs = []) ~label s = Code_block (attrs, label, s)
+
+let html_block ?(attrs = []) s = Html_block (attrs, s)
+
+let def_list ?(attrs = []) l = Definition_list (attrs, l)
+
 let of_channel ic = parse_inlines (Pre.of_channel ic)
 let of_string s = parse_inlines (Pre.of_string s)
 let to_html doc = Html.to_string (Html.of_doc doc)
--- a/src/omd.ml
+++ b/src/omd.ml
@@ -22,9 +22,9 @@ let strong ?(attrs = []) il = Strong (attrs, il)
 
 let code ?(attrs = []) s = Code (attrs, s)
 
-let hard_break attrs = Hard_break attrs
+let hard_break = Hard_break []
 
-let soft_break attrs = Soft_break attrs
+let soft_break = Soft_break []
 
 let link ?(attrs = []) li = Link (attrs, li)
 
@@ -41,7 +41,7 @@ let ol ?(attrs = []) ?(spacing = Loose) l =
 
 let blq ?(attrs = []) blocks = Blockquote (attrs, blocks)
 
-let hr attrs = Thematic_break attrs
+let hr = Thematic_break []
 
 let code_block ?(attrs = []) ~label s = Code_block (attrs, label, s)
 
--- a/src/omd.ml
+++ b/src/omd.ml
@@ -15,40 +15,25 @@ let parse_inlines (md, defs) =
   List.map (Mapper.map (parse_inline defs)) md
 
 let txt ?(attrs = []) s = Text (attrs, s)
-
 let em ?(attrs = []) il = Emph (attrs, il)
-
 let strong ?(attrs = []) il = Strong (attrs, il)
-
 let code ?(attrs = []) s = Code (attrs, s)
-
 let hard_break = Hard_break []
-
 let soft_break = Soft_break []
-
 let link ?(attrs = []) li = Link (attrs, li)
-
 let img ?(attrs = []) li = Image (attrs, li)
-
 let html ?(attrs = []) s = Html (attrs, s)
-
 let pg ?(attrs = []) il = Paragraph (attrs, il)
-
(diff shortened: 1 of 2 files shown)


src/omd.ml
```diff
@@ -14,6 +14,41 @@ let parse_inlines (md, defs) =
...
+
+let em ?(attrs = []) il = Emph (attrs, il)
+
+let strong ?(attrs = []) il = Strong (attrs, il)
+
+let code ?(attrs = []) s = Code (attrs, s)
+
+let hard_break attrs = Hard_break attrs
+
+let soft_break attrs = Soft_break attrs
```
tatchi, line 27: Would that be better to have one `br` constructor with a `soft` argument? Something like:

```ocaml
val br : ?soft:bool -> attributes -> attributes inline
```

src/omd.mli
```diff
@@ -46,6 +46,48 @@ type 'attr block =
...
+
+val em : ?attrs:attributes -> attributes inline -> attributes inline
+
+val strong : ?attrs:attributes -> attributes inline -> attributes inline
+
+val code : ?attrs:attributes -> string -> attributes inline
+
+val hard_break : attributes -> attributes inline
+
+val soft_break : attributes -> attributes inline
```
tatchi, line 59: Would that be better to have a single `br` constructor with a `soft` argument? Something like:

```ocaml
val br : ?soft:bool -> attributes -> attributes inline
```

src/omd.ml
shonfeder, line 27: So long as these constructors live in the `Omd` module (rather than the lower-level `Ast` module), I *think* we'd never have any need for attributes on the breaks? So perhaps these should just be:

```suggestion
let hard_break = Hard_break []

let soft_break = Soft_break []
```

What do you think?

src/omd.ml
```diff
@@ -14,6 +14,41 @@ let parse_inlines (md, defs) =
...
+let pg ?(attrs = []) il = Paragraph (attrs, il)
+
+let ul ?(attrs = []) ?(spacing = Loose) l = List (attrs, Bullet '-', spacing, l)
+
+let ol ?(attrs = []) ?(spacing = Loose) l =
+  List (attrs, Ordered (1, '.'), spacing, l)
+
+let blq ?(attrs = []) blocks = Blockquote (attrs, blocks)
+
+let hr attrs = Thematic_break attrs
```
shonfeder, line 44: Same thought here regarding attributes. Perhaps we can simplify this with:

```suggestion
let hr = Thematic_break []
```

src/omd.mli
shonfeder: I think it makes sense to have to separate constructs for this, matching the AST. It doesn't seem to cost anything in terms of maintenance or readability, and it means the user has to make less mappings in their head :)

shonfeder (commented):
A warm welcome and many thanks for the contributions, @tatchi!  I'm very glad you're pitching to help here :smile: 

LGTM, modulo one suggestion about elements that can't (afak) meaningfully do anything with attributes at this level.

Let me know what you think!

src/omd.mli
tatchi: Valid points, thanks for the suggestion 🤗

src/omd.ml
```diff
@@ -14,6 +14,26 @@ let parse_inlines (md, defs) =
...
+let ul ?(attrs = []) ?(spacing = Loose) l = List (attrs, Bullet '-', spacing, l)
+
+let ol ?(attrs = []) ?(spacing = Loose) l =
+  List (attrs, Ordered (1, '.'), spacing, l)
+
+let blq ?(attrs = []) blocks = Blockquote (attrs, blocks)
+let hr = Thematic_break []
+let code_block ?(attrs = []) ~label s = Code_block (attrs, label, s)
+let html_block ?(attrs = []) s = Html_block (attrs, s)
+let def_list ?(attrs = []) l = Definition_list (attrs, l)
```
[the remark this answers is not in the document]
tatchi: I think I like to have short names for the function constructors. Option 2 is quite invasive so I would go for option 3. 

When possible, we can try to match the Tyxml API? We can take some inspiration from: https://github.com/ocaml/omd/pull/211/files#diff-16f58dc281193a8336f88df2eeb14760638cfc1ce4a2a1c479c509e33f39738f

And for the rest, well, we can try to do our best 😅 

That would give something like: 
```diff
diff --git a/src/omd.ml b/src/omd.ml
index 012604c..8922228 100644
--- a/src/omd.ml
+++ b/src/omd.ml
@@ -18,22 +18,22 @@ let txt ?(attrs = []) s = Text (attrs, s)
 let em ?(attrs = []) il = Emph (attrs, il)
 let strong ?(attrs = []) il = Strong (attrs, il)
 let code ?(attrs = []) s = Code (attrs, s)
-let hard_break = Hard_break []
-let soft_break = Soft_break []
-let link ?(attrs = []) li = Link (attrs, li)
+let hard_br = Hard_break []
+let soft_br = Soft_break []
+let a ?(attrs = []) li = Link (attrs, li)
 let img ?(attrs = []) li = Image (attrs, li)
 let html ?(attrs = []) s = Html (attrs, s)
-let pg ?(attrs = []) il = Paragraph (attrs, il)
+let p ?(attrs = []) il = Paragraph (attrs, il)
 let ul ?(attrs = []) ?(spacing = Loose) l = List (attrs, Bullet '-', spacing, l)
 
 let ol ?(attrs = []) ?(spacing = Loose) l =
   List (attrs, Ordered (1, '.'), spacing, l)
 
-let blq ?(attrs = []) blocks = Blockquote (attrs, blocks)
+let blockquote ?(attrs = []) blocks = Blockquote (attrs, blocks)
 let hr = Thematic_break []
-let code_block ?(attrs = []) ~label s = Code_block (attrs, label, s)
-let html_block ?(attrs = []) s = Html_block (attrs, s)
-let def_list ?(attrs = []) l = Definition_list (attrs, l)
+let code_bl ?(attrs = []) ~label s = Code_block (attrs, label, s)
+let html_bl ?(attrs = []) s = Html_block (attrs, s)
+let dl ?(attrs = []) l = Definition_list (attrs, l)
 let of_channel ic = parse_inlines (Pre.of_channel ic)
 let of_string s = parse_inlines (Pre.of_string s)
 let to_html doc = Html.to_string (Html.of_doc doc)
```

`blockquote` feels a bit different to me since that's the only one that isn't abbreviated.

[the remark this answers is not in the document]
shonfeder: Good thinking to look to TyXml! 

I think blockquote is ok, despite the disparity, since it matches the html tag and the TyXml constructor.

These generally look ok to me. My only point of concern is around the `a` function. The markdown links seems to not really map exactly onto to general html anchors, but rather be a special case of them. But maybe I am over thinking that a bit.

Lastly, since we are not tending towards the side of making the function names match the value constructors perfectly, what would you think about this for the breaks?

```diff
-let hard_break = Hard_break []
-let soft_break = Soft_break []
+let br = Hard_break []
+let nl = Soft_break []
```

I think this matches the semantics of these things fairly well, and `br` fits the TyXml and HTML I think?

WDYT?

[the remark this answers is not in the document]
shonfeder: Thanks so much for the PR and sorry for the very loooong turn around in merging this in. :)

shonfeder (approved):
LGTM!

The only obvious improvement I can think of is in regards to API docs for the functions, but that's something we're lacking in general, and we can add that for these new helpers in a followup.

shonfeder:
I've reported the dependency build failure blocking ci here upstream: https://github.com/ocaml/camlp-streams/issues/7

PR #266, opened 2022-04-28
Update spec to 0.30

SquidDev:
This also refactors how entities are parsed, meaning there's only one function to do it (rather than two). I'm not 100% with the changes I've made to the parser combinator and `Sub` modules - happy to discuss those.

Issue #244, opened 2021-05-29
failing verification against emphasis parsing specs


shonfeder:
Failing specification units:

- [ ] 410
- [ ] 411
- [ ] 414
- [ ] 415
- [ ] 416
- [ ] 428
- [ ] 468
- [ ] 469

```diff
Done: 0/0 (jobs: 1)File "tests/spec-410.html", line 1, characters 0-0:
File "tests/spec-411.html", line 1, characters 0-0:
File "tests/spec-414.html", line 1, characters 0-0:
File "tests/spec-415.html", line 1, characters 0-0:
File "tests/spec-416.html", line 1, characters 0-0:
File "tests/spec-428.html", line 1, characters 0-0:
File "tests/spec-468.html", line 1, characters 0-0:
File "tests/spec-469.html", line 1, characters 0-0:
         git (internal) (exit 1)
(cd _build/default && /usr/bin/git --no-pager diff --no-index --color=always -u tests/spec-410.html tests/spec-410.html.new)
diff --git a/tests/spec-410.html b/tests/spec-410.html.new
index 85ec7d5..1f7dbb7 100644
--- a/tests/spec-410.html
+++ b/tests/spec-410.html.new
@@ -1 +1 @@
-<p><em>foo<strong>bar</strong>baz</em></p>
+<p><em>foo</em><em>bar</em><em>baz</em></p>
         git (internal) (exit 1)
(cd _build/default && /usr/bin/git --no-pager diff --no-index --color=always -u tests/spec-411.html tests/spec-411.html.new)
diff --git a/tests/spec-411.html b/tests/spec-411.html.new
index 1166aff..ea468e5 100644
--- a/tests/spec-411.html
+++ b/tests/spec-411.html.new
@@ -1 +1 @@
-<p><em>foo**bar</em></p>
+<p><em>foo</em><em>bar</em></p>
         git (internal) (exit 1)
(cd _build/default && /usr/bin/git --no-pager diff --no-index --color=always -u tests/spec-414.html tests/spec-414.html.new)
diff --git a/tests/spec-414.html b/tests/spec-414.html.new
index ef1a772..44259d7 100644
--- a/tests/spec-414.html
+++ b/tests/spec-414.html.new
@@ -1 +1 @@
-<p><em>foo<strong>bar</strong></em></p>
+<p><em>foo</em><em>bar</em>**</p>
         git (internal) (exit 1)
[...]

PR #269, opened 2022-05-13
fix some emphasis parsing

tatchi:
Not ready for review yet. Opening it as a draft to give an overview of what I'm currently working on.


tatchi:
Okay, I think I managed to fix all the broken tests mentioned in #244. At least tests are passing now and nothing else broke 😁

It's the first time I'm working on a parser, and that was not easy 🥵😅  

Based on the expected outcome, I first tried to guess what should be the solution until I told myself that it should probably be a good idea to read the spec. First time that I'm reading a spec (or rather a [part of it](https://spec.commonmark.org/0.30/#can-open-emphasis)), and that was very helpful! 😁

There are 3 main commits:

1. https://github.com/ocaml/omd/pull/269/commits/b52495aff1004bdc7eb2ce52a779fa0b9014591e that fixes **410,411, 414 and 428**
2. https://github.com/ocaml/omd/pull/269/commits/cceeb2c0b362fe1c3aee76345d8d455d0d9a2ce8 that fixes **415 and 416**
3. https://github.com/ocaml/omd/pull/269/commits/7e6a712d473a2453f5f70333cf3ff9eb3c3c2a4e that fixes **468 and 469**

I'm not very happy with the code of 1 and 3 so I'll see if I come up with something better. I'm planning on creating a separate PR for each of them so we can have a dedicated discussion for each. The fix is different anyway so I think it makes sense.

sonologico:
This is great. Thank you. Splitting PRs would be good.

shonfeder:
Yeah, this is really great work!

I agree that refactoring it into independent PRs could help keep things very clean.

One heads up is should we be aware of the potential with conflicts with #266. I'll work on trying to get that merged in the next day or two, which should help clear up that risk one way or another :)

shonfeder:
I would be fine with review this from the current PR. While I appreciate the intention to keep the PRs clean and focused, this group of spec violations was already grouped into a single issue, and the changeset is small enough here, and close enough related, that I think it makes sense to review in one PR.

tests/extract_tests.ml
```diff
@@ -8,8 +8,7 @@ let protect ~finally f =
       finally ();
       r
 
-let disabled =
-  [ 175; 184; 185; 410; 411; 414; 415; 416; 428; 468; 469; 516; 536 ]
+let disabled = [ 175; 184; 185; 516; 536 ]
```
shonfeder, line 11: This is where the only merge conflict is arising! Let me know if you have any trouble resolving the conflict here.

src/parser.ml
```diff
@@ -967,28 +967,61 @@ module Pre = struct
...
+                && n1 mod 3 != 0
+                && n2 mod 3 != 0
+              then
+                let rec find_next_emph = function
+                  | Emph (_, _, _, n) :: _ -> Some n
+                  | _ :: xs -> find_next_emph xs
+                  | [] -> None
+                in
+                let next_emph = find_next_emph xs in
+                match next_emph with
```
shonfeder, line 984: I don't think the temp var helps here, so we can reduce info load with:

```suggestion
                match find_next_emph xs with
```

src/parser.ml
```diff
@@ -967,28 +967,61 @@ module Pre = struct
...
+                | None -> loop (x :: acc) xs
+                | Some n ->
+                    if (n + n2) mod 3 = 0 && n mod 3 != 0 && n2 mod 3 != 0 then
```
shonfeder, line 987: Seems to be the same condition as we hit on lines 974 to 976, I think that's a good sign this could be factored out into a named predicate.

src/parser.ml
```diff
@@ -967,28 +967,61 @@ module Pre = struct
...
+                  if n1 >= 2 && n2 >= 2 then R (Strong ([], il)) :: xs
+                  else R (Emph ([], il)) :: xs
+                in
+                let r =
+                  if n1 >= 2 && n2 >= 2 then
+                    if n1 > 2 then Emph (pre, Other, q1, n1 - 2) :: r else r
+                  else if n1 > 1 then Emph (pre, Punct, q1, n1 - 1) :: r
+                  else r
+                in
+                parse_emph r
```
shonfeder, line 1010: I know you didn't add this code, but if you happened to have reasoned out what it does and can help illuminate with some comments, that would be a big help. I find it pretty inscrutable! If not, not worries :)

src/parser.ml
```diff
@@ -967,28 +967,61 @@ module Pre = struct
...
-                else R (Emph ([], il)) :: xs
-              in
-              let r =
-                if n1 >= 2 && n2 >= 2 then
-                  if n1 > 2 then Emph (pre, Punct, q1, n1 - 2) :: r else r
-                else if n1 > 1 then Emph (pre, Punct, q1, n1 - 1) :: r
-                else r
+              let next_closer_emph = find_next_closer_emph xs1 in
+              let is_next_closer_same =
+                match next_closer_emph with None -> false | Some q3 -> q2 = q3
```
shonfeder, line 1019: I appreciate using this temp variable to communicate the meaning here!

src/parser.ml
```diff
@@ -967,28 +967,61 @@ module Pre = struct
...
+                in
+                let r =
+                  if n1 >= 2 && n2 >= 2 then
+                    if n1 > 2 then Emph (pre, Other, q1, n1 - 2) :: r else r
+                  else if n1 > 1 then Emph (pre, Punct, q1, n1 - 1) :: r
+                  else r
+                in
+                parse_emph r
+          | (Emph (_, _, q2, _) as x) :: xs1 as xs when is_opener x ->
+              let rec find_next_closer_emph = function
```
shonfeder, line 1012: I think we can also move this helper out of the main program logic, and instead define it under the main function head (so move it up around line 968)?

src/parser.ml
shonfeder, line 1019: Using the same reasoning as the previous case, I suggest:

```suggestion
              let is_next_closer_same =
                match find_next_closer_emph xs1 with None -> false | Some q3 -> q2 = q3
```

src/parser.ml
```diff
@@ -967,28 +967,61 @@ module Pre = struct
...
+                match next_closer_emph with None -> false | Some q3 -> q2 = q3
               in
-              parse_emph r
-          | (Emph _ as x) :: xs1 as xs when is_opener x ->
-              let xs' = parse_emph xs in
-              if xs' = xs then loop (x :: acc) xs1 else loop acc xs'
+              if not is_next_closer_same then loop (x :: acc) xs1
+              else
+                let xs' = parse_emph xs in
+                if xs' = xs then loop (x :: acc) xs1 else loop acc xs'
```
shonfeder, line 1024: I haven't puzzled out this function enough, do you happen to know why we are doing a (polymorphic) equality comparison on the lists here

shonfeder (commented):
Left some initial questions and suggestions.

I think your additions generally read ok and fits with the existing code around it.

My biggest ask is that you provide comments to explain the logic you're introducing. This is something we are sorely lacking in the existing parsing code, and it would be awesome to start sharing the knowledge while it is fresh in people's minds after they have to reconstruct it from the code.

All of my other suggestions here are negotiable if you have strong feelings that point in another direction.

tatchi:
Thanks for the review! 😊 I would have preferred to split that PR into 3 smaller ones, but now that you started the review process, let's continue here :) 

I will address your comments and try to do my best to document my changes.

src/parser.ml
tatchi: I have no idea 😬 I could investigate but that comparison doesn't seem to be needed as the following code successfully pass all the tests.

```diff
diff --git a/src/parser.ml b/src/parser.ml
index e90b96d..46f37b6 100644
--- a/src/parser.ml
+++ b/src/parser.ml
@@ -959,9 +959,7 @@ module Pre = struct
                 | Some (_, _, q3, _) -> q2 = q3
               in
               if not is_next_closer_same then loop (x :: acc) xs1
-              else
-                let xs' = parse_emph xs in
-                if xs' = xs then loop (x :: acc) xs1 else loop acc xs'
+              else loop acc (parse_emph xs)
           | x :: xs -> loop (x :: acc) xs
           | [] -> x :: List.rev acc
         in

```

src/parser.ml
```diff
@@ -967,28 +967,61 @@ module Pre = struct
...
-                  if n2 > 2 then Emph (Punct, post, q2, n2 - 2) :: xs else xs
-                else if n2 > 1 then Emph (Punct, post, q2, n2 - 1) :: xs
-                else xs
+          | (Emph (_, post, q2, n2) as x) :: xs as xall
+            when is_closer x && q1 = q2 ->
+              if
+                is_opener x
+                && (n1 + n2) mod 3 = 0
+                && n1 mod 3 != 0
+                && n2 mod 3 != 0
```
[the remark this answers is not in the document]
tatchi: I added an `is_match` function with a description coming from the spec: https://github.com/ocaml/omd/pull/269/files#diff-2f56b48efe4c0b849c5b527b223a6c8ab85abd599e10f8b465f5c9801449a468R921

Not convinced by the function's name though but I couldn't come up with anything better. Open to any better alternative :)

tatchi:
I should have addressed your comments. I also added some comments to document the code I added. That wasn't super easy to explain so hopefully it's more or less understandable 😅

I'm wondering if the functions `find_next_emph` I introduced are ideal. Instead of looking at the "future" (i.e what we haven't read yet) to make a decision, I'm wondering if we couldn't continue to read in any case but store some info in the `acc` (i.e what was the last delimiter we read) to make a decision later on.

I'm not sure if that makes sense and if that's feasible. I'll try and see how far I can go.

src/parser.ml
```diff
@@ -904,31 +904,124 @@ module Pre = struct
...
+
+  let rec find_next_closer_emph = function
+    | (Emph (pre, post, style, n) as e) :: _ when is_closer e ->
+        Some (pre, post, style, n)
+    | _ :: xs -> find_next_closer_emph xs
+    | [] -> None
+
+  (* From the spec: "If one of the delimiters can both open and close emphasis, then the sum of the lengths
+      of the delimiter runs containing the opening and closing delimiters must not be
+      a multiple of 3 unless both lengths are multiples of 3" *)
```
shonfeder, line 920:
```suggestion
  (* Checks the lengths of two different emphasis delimiters to see if there can be a match
  
      From the spec: "If one of the delimiters can both open and close emphasis, then the sum of the lengths
      of the delimiter runs containing the opening and closing delimiters must not be
      a multiple of 3 unless both lengths are multiples of 3" *)
```

src/parser.ml
```diff
@@ -904,31 +904,124 @@ module Pre = struct
...
+                 according to the "mod 3" rule. In that case, we check if the next delimiter can match.
+                
+                   *foo**bar**baz*  The second delimiter that's both an opener/closer ( ** before bar)
+                                    matches with the next delimiter ( ** after bar). They'll become
+                                    <strong>bar</strong>. The end result will be: <em>foo<strong>bar</strong>baz</em>
+                
+                
+                   *foo**bar*baz*  The second delimiter that's both an opener/closer ( ** before bar)
+                                   doesn't match with the next delimiter ( * after bar). **bar will be
+                                   considered as regular text. The end result will be: <em>foo**bar</em>baz*
```
shonfeder, line 975: Dang, this is so subtle. Having the comment explaining really helps.

src/parser.ml
shonfeder: That's already a good investigation! I'll make note to followup here with a change that removes this bit. Thanks!

shonfeder (commented):
I have a few followup suggestions, but looks excellent to me.

Thank you very much for your careful work here. The comments are a huge help in understanding what's going on in this code and, moreover and more importantly, set a good example for the rest of us to follow as we continue maintaining the parser.

shonfeder:
>  I would have preferred to split that PR into 3 smaller ones, but now that you started the review process, let's continue here :)

I'm sorry about that! I didn't mean to push you into doing just one PR. I'll refrain from doing a review before you indicate you're ready in the future.

As soon as you are happy with this one, please convert it from a draft PR. All that's left to be done there is to add an entry into the changelog (I'll do that if you haven't beaten me to it by the time this is converted from a draft).

Thanks again for your excellent work here on fixing the parser! :)

shonfeder:
Hi, @tatchi! Just checking here. If you're happy with the state of things, I'd be happy to make a small add here to update the changelog and merge this in. But I don't want to step on your toes in case you had been planning to do more work on this.

tatchi:
> Hi, @tatchi! Just checking here. If you're happy with the state of things, I'd be happy to make a small add here to update the changelog and merge this in. But I don't want to step on your toes in case you had been planning to do more work on this.

Hi @shonfeder, everything looks good to me except these two `find_next_*` functions that I introduced. I'm actually not sure if that matters much, but I have the feeling that it would be better not to use such functions that look ahead but instead store info about what we've parsed so far to make future decisions.

I'm not sure if that will be feasible but I would like at least to give it a try. Good news is that I'll have time to look at it this week, so expect an outcome by the end of the week 😁

src/parser.ml
tatchi: I tried removing that equality comparison in master

```diff
diff --git a/src/parser.ml b/src/parser.ml
index 6342989..053ca50 100644
--- a/src/parser.ml
+++ b/src/parser.ml
@@ -926,9 +926,8 @@ module Pre = struct
                 else r
               in
               parse_emph r
-          | (Emph _ as x) :: xs1 as xs when is_opener x ->
-              let xs' = parse_emph xs in
-              if xs' = xs then loop (x :: acc) xs1 else loop acc xs'
+          | (Emph _ as x) :: _ as xs when is_opener x ->
+              loop acc (parse_emph xs)
           | x :: xs -> loop (x :: acc) xs
           | [] -> x :: List.rev acc
         in
```

When running the tests, it gets stuck at 99%. I assume it's because of an infinite loop. So it appears that this code is useful in master but becomes unnecessary with that PR.

tatchi:
I discovered an emphasis parsing bug that was not covered by the conformance tests. I added an extra test in https://github.com/ocaml/omd/pull/269/commits/46d6a612b497c8653548a803198d1f215cc6f628 with the fix in https://github.com/ocaml/omd/pull/269/commits/6e17dd3211cbe4b56bd4f7be9ea26a7c274e0e89

For the rest, I haven't found a way to remove these two `find_next_*` functions, but I don't think it's important. 

I marked the PR as ready for review so it can be merged now 😁

shonfeder:
Many thanks for the great work here, @tatchi!

PR #273, opened 2022-06-20
Port 1.3 to dune

tmattio:
This PR backports the switch of the build system from oasis to dune.

This is part of the OCaml release readiness effort for OCaml 5. In particular, omd is a dependency of `ocaml-lsp-server`, so the build failure of omd on OCaml 5 prevents us from using LSP there.

Omd 1.3.1 builds with Oasis, which isn't compatible with OCaml 5. Seeing that the project is deprecated in the [OCaml Platform](https://ocaml.org/docs/platform), I am not sure we'll be making it compatible. Instead, it seems sensible to backport Dune support to Omd to allow it to build on OCaml 5.

Some remarks on the PR:

- it might be more invasive than it needs. In particular, I chose to delete all the files that weren't necessary, rewrite the Makefile, add dune rules for tests, etc. I'm happy to revert any of these if you'd prefer a more minimal PR.
- One of the interface (`src/omd_lexer_fs.mli`) wasn't correct. I suspect that the file wasn't being built. This PR fixes the interface, but the correct change to make it backward compatible is to remove the file I think.


nojb:
Wouldn't it be better to put energy into making a release of the current code, rather than marching on with the antediluvian 1.3.1 release?

tmattio:
I certainly wouldn't mind a stable release of Omd 2, but I think we'd still want to make omd 1.3.1 compatible with OCaml 5. Here's a list of reverse dependencies:

- builder-web
- camyll
- cow
- cowabloga
- finch
- hardcaml-examples
- hardcaml-framework
- jekyll-format
- lambdoc
- learn-ocaml
- learn-ocaml-client
- malfunction
- md2mld
- monorobot
- ocaml-lsp-server
- ocp_reveal
- sail
- stone

I doubt we'll want to migrate all of these in the context of OCaml 5 release readiness.

It also seems like https://github.com/ocaml/omd/pull/269 is going to be necessary to cut a stable release. It's still in draft though, and I wouldn't want to rush the release of omd: we'd need the Platform tools to be compatible with OCaml 5 before alpha1.

shonfeder:
I agree that we should support 1.3. We still are working to achieve conformity to the spec, and the current state of 2.0 doesn't begin to approach the extensibility or features supplied by 1.3 afaik.

We could push a stable release of 2.0 even without fixing the bugs that cause us to deviate from the spec, but the maintenance burden to migrate all the dependencies would be very significant.

sonologico:
Agree on supporting 1.3 for now. I don't think we should do much more than this PR though.

* About `omd_lexer_fs`. I think you're right that it wasn't compiled at all.

sonologico (approved):

shonfeder (approved):

omd.opam
```diff
@@ -1,24 +1,32 @@
 # This file is generated by dune, edit dune-project instead
```
shonfeder, line 1: I didn't mean to commit this! Sorry :( 

Seems to indicate some inconsistency in our dune builds tho? I'm not sure what accounts for that.

shonfeder (approved):
Fixed up the test and added CI.

Thank you very much for the assist here, @tmattio! I'll cut a release to to the opam repo this week.

src/omd_html.ml
```diff
@@ -38,7 +38,7 @@ let to_string html =
         | (a, Some v) ->
           if not (String.contains v '\'') then
             pp " %s='%s'" a v
-          else not (String.contains v '"') then
+          else if not (String.contains v '"') then
```
shonfeder, line 41: This is curious. I'm not sure how this would have been compiling before!

shonfeder (approved):

PR #275, opened 2022-07-28
pin camlp-streams

tatchi:
I created this branch to verify that https://github.com/ocaml/camlp-streams/pull/8 fixes our build issue. I thought I would open a PR in case you would like to merge it as a temporary solution.

Feel free to close it if you prefer to wait for a proper solution.

### DIFF
--- a/omd.opam
+++ b/omd.opam
@@ -41,3 +41,6 @@ build: [
   ]
 ]
 dev-repo: "git+https://github.com/ocaml/omd.git"
+pin-depends: [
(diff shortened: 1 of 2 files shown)


shonfeder:
Thank you for trying this out and showing `ocamlp-streams` that this does indeed resolve the problem. For the purposes of our work on the new version of OMD, I think we can safely (and with less overhead) just bump our minimum supported ocaml version to 4.05 (also suggested in https://github.com/ocaml/camlp-streams/issues/7#issuecomment-1193986526). See #279

PR #276, opened 2022-07-30
refactor: use labelled arguments

tatchi:
Just because I find it easier to track what's happening in this recursive function with the labelled arguments. Without the labels, you always have to remember the order of these two booleans' arguments and what's their meaning.


shonfeder (approved):
LGTM. Will merge in once we get the CI sorted out. Thanks!

PR #277, opened 2022-07-31
unicode link label normalization (fix test 539)

tatchi:
Input:

```
[ẞ]

[SS]: /url
```

This is the result in master:

```diff
File "tests/spec-539.html", line 1, characters 0-0:
diff --git a/_build/default/tests/spec-539.html b/_build/default/tests/spec-539.html.new
index c31102f..4e28a6f 100644
--- a/_build/default/tests/spec-539.html
+++ b/_build/default/tests/spec-539.html.new
@@ -1 +1 @@
-<p><a href="/url">ẞ</a></p>
+<p>[ẞ]</p>
make: *** [test] Error 1   
```

The issue is that both labels are not being matched, hence is it not recognized as a link. To match labels, we need to `normalize` them (strip off leading/trailing whitespace, ...) and do a case-insensitive comparison. The unicode version of that is a bit more complex as we need to do a `Unicode case folding`. From the spec:

>  One label [matches](https://spec.commonmark.org/0.30/#matches) another just in case their normalized forms are equal. To normalize a label, strip off the opening and closing brackets, perform the Unicode case fold, strip leading and trailing spaces, tabs, and line endings, and collapse consecutive internal spaces, tabs, and line endings to a single space.

This PR adapts the `normalize` function to work with unicode labels too. Fortunately, I could rely on some libs (`uutf`, `uucp`, and `uunf`) and I even found a [piece of code in the doc](https://erratique.ch/software/uucp/doc/Uucp/Case/index.html#caselesseq) that does almost what's needed.

With that adapted `normalize` function, `ẞ` and `SS` are matched. The result is now a link as expected.

### DIFF
--- a/src/parser.ml
+++ b/src/parser.ml
@@ -1100,7 +1100,7 @@ let normalize s =
       in
       add
     in
-    let add =
+    let add_nfd =
       let n = Uunf.create `NFD in
       let rec add v =
         match Uunf.add n v with
@@ -1113,26 +1113,20 @@ let normalize s =
       in
       add
     in
-    let fold () =
-      let start = ref true in
-      let seen_ws = ref false in
-      let add_uchar () _ = function
-        | `Malformed _ ->
-            start := false;
-            seen_ws := false;
-            add (`Uchar Uutf.u_rep)
-        | `Uchar u as uchar ->
-            if Uucp.White.is_white_space u then seen_ws := true
-            else (
-              if (not !start) && !seen_ws then add (`Uchar (Uchar.of_char ' '));
-              add uchar;
-              start := false;
-              seen_ws := false)
-      in
-      add_uchar
+    let uspace = `Uchar (Uchar.of_char ' ') in
+    let add_uchar (start, seen_ws) _ = function
+      | `Malformed _ ->
+          add_nfd (`Uchar Uutf.u_rep);
+          (false, false)
+      | `Uchar u as uchar ->
+          if Uucp.White.is_white_space u then (start, true)
+          else (
+            if (not start) && seen_ws then add_nfd uspace;
+            add_nfd uchar;
+            (false, false))
     in
-    Uutf.String.fold_utf_8 (fold ()) () s;
-    add `End;
+    let (_ : bool * bool) = Uutf.String.fold_utf_8 add_uchar (true, false) s in
+    add_nfd `End;
     to_nfd_and_utf_8 `End;
     Buffer.contents b
   in
--- a/src/parser.ml
+++ b/src/parser.ml
@@ -1085,19 +1085,58 @@ let link_label allow_balanced_brackets st =
   in
   loop 0 false
 
+(* based on https://erratique.ch/software/uucp/doc/Uucp/Case/index.html#caselesseq *)
 let normalize s =
-  let buf = Buffer.create (String.length s) in
-  let rec loop ~start ~seen_ws i =
-    if i >= String.length s then Buffer.contents buf
-    else
-      match s.[i] with
-      | ' ' | '\t' | '\010' .. '\013' -> loop ~start ~seen_ws:true (succ i)
-      | _ as c ->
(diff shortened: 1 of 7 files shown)


src/parser.ml
```diff
@@ -989,19 +989,58 @@ let link_label allow_balanced_brackets st =
...
+      let rec add v =
+        match Uunf.add n v with
+        | `Await | `End -> ()
+        | `Uchar u ->
+            (match Uucp.Case.Fold.fold u with
+            | `Self -> to_nfd_and_utf_8 (`Uchar u)
+            | `Uchars us -> List.iter (fun u -> to_nfd_and_utf_8 (`Uchar u)) us);
+            add `Await
+      in
+      add
```
shonfeder, line 1018: With all the `add`s flying around here, I am prone to get a bit confused. Perhaps it would help clarify for readers like me to give this outer `add` a more specific name? Perhaps like `add_nfd`?

src/parser.ml
```diff
@@ -989,19 +989,58 @@ let link_label allow_balanced_brackets st =
...
+      let seen_ws = ref false in
+      let add_uchar () _ = function
+        | `Malformed _ ->
+            start := false;
+            seen_ws := false;
+            add (`Uchar Uutf.u_rep)
+        | `Uchar u as uchar ->
+            if Uucp.White.is_white_space u then seen_ws := true
+            else (
+              if (not !start) && !seen_ws then add (`Uchar (Uchar.of_char ' '));
```
shonfeder, line 1031: I'm a bit unsure whether using mutable state here is producing the most readable (or performant) fold. I've a suspicion we could use the accumulator arg of the folded function to track the state of whether we've seen white space, and I'm not sure convinced that `start` is needed at all.

I'll see if I can sketch out what I have in mind later this week, unless you point out something essential I'm missing :)

shonfeder (commented):
Very nice improvement and fix! I've two minor suggestions. I'll stage my suggestions (and perhaps merge conflict cleanup) as PR into this PR, unless you prefer (and have time) to address those things before I can get to it :D

src/parser.ml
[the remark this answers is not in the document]
tatchi: Definitely makes sense. I thought about it when I initially wrote that code and then forgot it 🙈

tatchi: That's what I did in the first version I implemented but since the return value is not used I thought perhaps it's just fine to use mutation.

I made a version in https://github.com/ocaml/omd/pull/277/commits/4e96339463976f4fd1d7e308180a4a85c973b42d that uses a tuple of two boolean to track the `start` and `seen_ws` values. I'm not sure that's ideal since we can easily mix them, maybe a record instead is better?

tatchi: See https://github.com/ocaml/omd/pull/277/commits/e51ad122add893419765416e6876f2a54d21b27f for the version that uses record. Let me know which one you prefer

shonfeder: I see what you mean. I had the false impression we could do away with `start`. Thank you for explaining the need. Given that requirement, I don't have a preference over the record vs. the mutable refs and I'm happy to go with whichever you prefer here.

Sorry for the red herring here!

shonfeder (approved):
LGTM!

tatchi:
Thanks for the review @shonfeder, I added extra tests to cover the normalization of labels in https://github.com/ocaml/omd/pull/277/commits/49d3ca2a12583d0804052ab40406b916081a7ebc

tests/extract_tests.ml
```diff
@@ -8,7 +8,7 @@ let protect ~finally f =
       finally ();
       r
 
-let disabled = [ 206; 215; 216; 519; 539 ]
+let disabled = [ 206; 215; 216; 519 ]
```
shonfeder, line 11: Outstanding progress. You're demolishing these deviations. Thanks so much @tatchi! :tada:

shonfeder (approved):

PR #283, opened 2022-08-11
fix 215 and 216

tatchi:
### Test 215

```md
[foo]: /url
bar
===
[foo]
```

Current result:

```
<h1>[foo]: /url                     
bar</h1>
<p>[foo]</p>
```

Expected result:

```
<h1>bar</h1>
<p><a href="/url">foo</a></p>
```

The issue is that when parsing a `Lsetext_heading` (===), we were wrapping all the `Rparagraph` lines in a `Heading`, without taking into account that some of the `lines` might be a `link reference definition`

### Test 216

```
[foo]: /url
===
[foo]
```

Current result:

```
<h1>[foo]: /url</h1>                
<p>[foo]</p>
```

Expected result:

```
<p>===
<a href="/url">foo</a></p>
```

Same as 216 except that in that case there's nothing between the `link reference definition` and the `Lsetext_heading`. There's nothing then to make as `Heading`.

### DIFF
--- a/tests/extract_tests.ml
+++ b/tests/extract_tests.ml
@@ -8,7 +8,7 @@ let protect ~finally f =
       finally ();
       r
 
-let disabled = [ 215; 216; 519 ]
+let disabled = [ 519 ]
 
 let with_open_in fn f =
   let ic = open_in fn in
--- a/src/block.ml
+++ b/src/block.ml
@@ -39,6 +39,11 @@ module Pre = struct
     let i = loop 0 in
     if i > 0 then String.sub s i (String.length s - i) else s
 
+  let link_reference_definitions s =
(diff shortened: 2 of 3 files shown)


tests/extract_tests.ml
```diff
@@ -8,7 +8,7 @@ let protect ~finally f =
       finally ();
       r
 
-let disabled = [ 215; 216; 519 ]
```
tatchi, line 11: 💪💪

tatchi: To be fair though, I already gave a try at fixing the remaining one but it definitely doesn't feel easy 😬. I'll continue to investigate but I can't promise anything 😅

sonologico (approved):
Thank you!

PR #284, opened 2022-08-25
fix 519

tatchi:
This fixes the remaining failing test (519).

### DIFF
--- a/src/parser.ml
+++ b/src/parser.ml
@@ -1719,13 +1719,13 @@ let rec inline defs st =
     | exception Fail ->
         junk st;
         loop (Left_bracket kind :: text acc) st
-  and loop acc st =
+  and loop ~seen_link acc st =
     match peek_exn st with
     | '<' as c -> (
         match protect autolink st with
         | def ->
             let attr = inline_attribute_string st in
-            loop (Pre.R (Link (attr, def)) :: text acc) st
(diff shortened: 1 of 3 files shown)


src/parser.ml
```diff
@@ -1719,13 +1719,13 @@ let rec inline defs st =
     | exception Fail ->
         junk st;
         loop (Left_bracket kind :: text acc) st
-  and loop acc st =
+  and loop ~seen_link acc st =
```
tatchi, line 1722: We can avoid passing this `seen_link` argument every time by storing of the `Url` delimiter `[` or `]` is active or not. That's what I did in this branch: https://github.com/ocaml/omd/compare/master...fix-519

tatchi: Let me know which solution do you prefer :)

shonfeder: I'm ok with this approach! I assume you also prefer it since it's what you used for the PR :)

shonfeder (approved):

shonfeder:
Amazing. Thank you for the wonderful work dispatching all the spec exceptions, @tatchi!

I think one thing we've omitted in the last PR's is an update to the changelog that records these fixes, and celebrates your responsibility for them :)

If you'd like to add a PR memorializing your fixes in a changelog entry, please do. Otherwise I'll take care of adding that this weekend. 

:tada:

PR #286, opened 2022-09-03
Initial cleanup and refactoring of the parser

shonfeder:
## Context 

To help clear the way for #223 we've agreed that some substantial refactoring of the parer passes should be undertaken to make it less unwieldy to extend and reason about (following the principle of "first make the change easy, then make the change").

This work is largely just preparatory, and the result of "active reading", which is adding comments, cleanup, and simplification as I study the existing parser. I'll resume this work later in the week.

## Reviewing

There was some churn in the changes, but review still may benefit from skimming each commit, before a final review, to see the evolution of the changes.

I'd particularly like the review to ensure that the changes I'm proposing actually make the code more readable and easier to reason about. I don't want to be introducing changes for their own sake, so if you think any of my changes end up just as complicated or hard to reason about as what they are replacing (including the naming suggestions), please raise alarm!

Thanks in advance for the review!


src/parser.ml
```diff
@@ -82,9 +98,13 @@ end = struct
     in
     loop n s
 
-  let tails n { base; off; len } =
+  let drop n s =
     if n < 0 then invalid_arg "tails";
```
tatchi, line 102:
```suggestion
    if n < 0 then invalid_arg "drop";
```

shonfeder: Thanks!

shonfeder:
It's a long weekend in Canada, so I'll keep doing the cleanup here thru tomorrow, then I'll offer up whatever I have done by EOD for reviw :)

shonfeder:
That's what I'll be getting done this weekend! Merging will be blocked until I sort out https://github.com/thierry-martinez/stdcompat/issues/26, but this should be ready for review (mainly just to keep it from getting too long).

I'll resume this superficial pass of parsing later in the week, or on the weekend at the latest. I think I'm forming some ideas for more substantial refactoring, and I'll open issues to plan those out as they become more formed.

shonfeder:
Interesting! It looks like some of our custom `Compat` functions are compatible with 4.14, but break behavior related to unicode on older versions of OCaml! I'll have to restore our `Compat`, probably as an overlay on top of `Stdcompat`, and just for whichever function is doing the magic here.

src/block.ml
```diff
@@ -1,5 +1,5 @@
 open Ast
-module Sub = Parser.Sub
+module Sub = StrSlice
```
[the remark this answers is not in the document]
sonologico: I find `StrSlice` a better name than `Parser.Sub`.

src/parser.ml
```diff
@@ -139,24 +100,23 @@ end = struct
 
   type 'a t = state -> 'a
 
+  let ensure_chars_remain st = if st.pos >= String.length st.str then raise Fail
```
sonologico, line 103: Nitpick, but `ensure_chars_remain` to me sounds like it prevents chars from being removed whereas it is more like "ensure there are still chars". Can't think of an actual good name. Admittedly not great suggestions:
* `ensure_remaining_chars`
* `ensure_not_at_end`
* `ensure_end_not_reached`
* `fail_if_no_more_chars`

src/parser.ml
```diff
@@ -237,97 +194,99 @@ type t =
   | Lthematic_break
   | Latx_heading of int * string * attributes
   | Lsetext_heading of int * int
+      (** the level of the heading and how long the underline marker is *)
```
sonologico, line 197: Should we consider inline records here if they don't make later code more verbose?

src/parser.ml
shonfeder: I can see the potential for confusion. Since it's only being used in 2 places, maybe best just to remove this? It also lets us make avoid conditional statements and use conditional expression. WDYT? https://github.com/ocaml/omd/pull/286/commits/6611e256f7a5090a80f62ed5ea691a1afa9bb44d

src/block.ml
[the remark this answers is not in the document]
shonfeder: Cool. I think I do too. I dropped the `Sub` alias.

src/parser.ml
shonfeder: Good idea!

tatchi:
> Interesting! It looks like some of our custom `Compat` functions are compatible with 4.14, but break behavior related to unicode on older versions of OCaml! I'll have to restore our `Compat`, probably as an overlay on top of `Stdcompat`, and just for whichever function is doing the magic here.

Another option could be to use the `Uutf.Buffer.add_utf_8` function instead (and get rid of the `Compat` module). See [my commit](https://github.com/tatchi/omd/pull/1/commits/e1f1f94834f4cec1fcb424113fc50d7c9266da40#diff-2f56b48efe4c0b849c5b527b223a6c8ab85abd599e10f8b465f5c9801449a468R464)

shonfeder:
That's perfect. Thanks for the fix, @tatchi!

dune-project
```diff
@@ -24,7 +24,8 @@ Additionally, OMD implements a few Github markdown features, an
 extension mechanism, and some other features. Note that the opam
 package installs both the OMD library and the command line tool `omd`.")
  (tags (org:ocamllabs org:mirage))
- (depends (ocaml (>= 4.05))
+ (depends (ocaml (>= 4.08))
```
shonfeder, line 27: Note that I'm suggesting raising the minimum ocaml version here. I think this library is high-level enough, and has few enough dependencies that we don't need to weigh ourselves down trying to maintain backwards compatibility with versions that much more central libraries have already lost support for.

shonfeder:
Thanks to @tatchi's suggestion and my suggestion to increase our minimum OCaml version to 4.08 (for the time being at least) I think this should be passing all CI now and ready for another review.

tatchi (approved):
Nice refactor 👏 Thanks for doing that!

PR #288, opened 2022-09-12
Rework element constructors

shonfeder:
Followup to #268.

These changes expand and rework the constructors for AST values proposed 
in #252. The aim is to make construction of AST values relatively simple
and light weight.

The constructors lacked some ergonomic convenience, and we were missing
a constructor for headings and missing a way to construct elements that
included concatenated inlines (so, e.g., there was no way of constructing
a paragraph that would include mixing plain text with `em` or `strong` text.

I think this was mostly due to us having omitted  tests that
would show how the constructors were meant to be used when we first
introduced them, so I'm adding in tests as I go so we can see how usage
will look, and spot any irregularities or oversights.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -27,7 +27,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Use OCaml ${{ matrix.ocaml-compiler }}
-        uses: ocaml/setup-ocaml@v2.0.3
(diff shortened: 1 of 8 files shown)


shonfeder:
This presupposes #289 and I'll rebase on master once that is merged in, before requesting a review.

tatchi (approved):
LGTM! Thanks for adding the tests and the doc 👌

PR #290, opened 2022-10-13
Remove obsolete documentation

cuihtlauac:
It seems installing dependencies using make no longer makes sense


shonfeder:
Since we have the make target for this anyhow, I don't think there's any harm in leaving the docs for the make target in the readme.

Issue #297, opened 2022-12-06
Empty Inline


cuihtlauac:
Wherever an inline value is required, it could actually be “nothing”. It would be helpful to have an extra constructor in the inline type that would represent that. As it is today, inline feels like an unlucky monoid which lost its neutral element :-)

cuihtlauac:
PR: https://github.com/ocaml/omd/pull/298

PR #298, opened 2022-12-06
Add Empty inline

cuihtlauac:
Fix https://github.com/ocaml/omd/issues/297

### DIFF
--- a/src/ast_inline.ml
+++ b/src/ast_inline.ml
@@ -6,6 +6,7 @@
         but with the caveat that most renderings of the document don't support
         attributes in these nodes? *)
 type 'attr inline =
+  | Empty
   | Concat of 'attr * 'attr inline list
   | Text of 'attr * string
   | Emph of 'attr * 'attr inline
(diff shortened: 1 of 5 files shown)


src/ast_inline.ml
```diff
@@ -6,6 +6,7 @@
         but with the caveat that most renderings of the document don't support
         attributes in these nodes? *)
 type 'attr inline =
+  | Empty
   | Concat of 'attr * 'attr inline list
```
shonfeder, line 10: I'm slightly unsure what we gain by introducing the `Empty` data constructor here. If we want a unit value, that would already be available as `Concat ([], [])`, and we could add, e.g., a named value as a shorthand for this: `let empty attr = Concat (attr, [])`

Meanwhile, adding a new data constructor will require additional cases during analysis that don't actually improve the logic afaict.

Do you perhaps have any examples in mind where an explicit `Empty` data constructor is helps clarify things that couldn't be addressed through my proposed `empty` value?

cuihtlauac: Sorry, I should have explained where this comes from.

The initial use case came from the need being able to have a link with an empty label (of type inline), filled later with CSS.

I fully agree using a shorthand, as you propose, is a better solution, as it avoids having two ways to represent the same thing (Empty and Concat ([], []). I had overlooked that.

I'll update the PR with your proposal.

shonfeder (approved):
Looks great. Thanks!

Issue #299, opened 2023-01-03
Expose function Html.to_plain_text


cuihtlauac:
In order to add `id` attributes, `ocaml/ocaml.org` has duplicated the function `Html.to_plain_text`. See https://github.com/ocaml/ocaml.org/blob/main/tool/ood-gen/lib/tutorial.ml#L28

Therefore, it seems to make sense to expose this function publicly.

PR #300, opened 2023-01-03
Expose Html.to_plain_text as Omd.to_string

cuihtlauac:
Expose function allowing to turn an inline element into a string,
keeping only the text.

This is fix for issue https://github.com/ocaml/omd/issues/299

### DIFF
--- a/src/omd.ml
+++ b/src/omd.ml
@@ -32,3 +32,5 @@ let to_html ?auto_identifiers doc =
   Html.to_string (Html.of_doc ?auto_identifiers doc)
 
 let to_sexp ast = Format.asprintf "@[%a@]@." Sexp.print (Sexp.create ast)
+
+let to_string = Html.to_plain_text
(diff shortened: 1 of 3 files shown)


src/omd.mli
```diff
@@ -30,3 +30,6 @@ val of_channel : in_channel -> doc
 val of_string : string -> doc
 val to_html : ?auto_identifiers:bool -> doc -> string
 val to_sexp : doc -> string
+
+val to_string : 'attr inline -> string
```
shonfeder, line 34: Perhaps this should be
 
```suggestion
val to_plain_text : 'attr inline -> string
```

to avoid confusion?

shonfeder (commented):
Thanks!

What do you think of my alternative naming suggestion?

cuihtlauac:
It is much better indeed. Can you do it? I'm afk this week.

src/omd.ml
```diff
@@ -32,3 +32,4 @@ let to_html ?auto_identifiers doc =
   Html.to_string (Html.of_doc ?auto_identifiers doc)
 
 let to_sexp ast = Format.asprintf "@[%a@]@." Sexp.print (Sexp.create ast)
+let to_string = Html.to_plain_text
```
shonfeder, line 35:
```suggestion
let to_plain_text = Html.to_plain_text
```

PR #301, opened 2023-02-10
move auto id into AST

### LINKED ISSUE #296: move auto generated id into AST
See https://github.com/ocaml/omd/pull/267#discussion_r973599653

tatchi:
Fixes #296 

This moves the auto identifiers logic into the AST (when parsing the document). It moves the `auto_identifiers` parameter from the `Html.of_doc` function to the `Omd.{of_channel,of_string}` function.

I'm not sure if this API is ideal because it means that we have to parse a document twice (once with `auto_identifiers` set to `true` and once with `false`) if we want to print it with/without auto-identifiers.

```ocaml
let with_ids = Omd.of_string ~auto_identifiers:true "..." |> Html.of_doc |> Html.to_string
let without_ids = Omd.of_string ~auto_identifiers:false "..." |> Html.of_doc |> Html.to_string
```

Perhaps including the auto-ids in the AST and deciding whether or not to include them at the time of printing would be a better solution.

```ocaml
let doc = Omd.of_string "..." |> Html.of_doc in
let with_ids = Html.to_string ~auto_identifiers:true doc
let without_ids = Html.to_string ~auto_identifiers:false doc
```

The only problem is that we'd need to distinguish an explicit id from an auto-generated one in the AST.

### DIFF
--- a/tests/expect_tests.ml
+++ b/tests/expect_tests.ml
@@ -34,10 +34,10 @@ let%expect_test "construct inline elements" =
 
 let%expect_test "construct headings" =
   show Omd.Ctor.[ h 1 ~attrs:[ ("class", "my-class") ] [ txt "Heading 1" ] ];
-  [%expect {| <h1 id="heading-1" class="my-class">Heading 1</h1> |}];
+  [%expect {| <h1 class="my-class">Heading 1</h1> |}];
   show Omd.Ctor.[ h 6 [ txt "Heading 6"; em "with emphasis!" ] ];
   [%expect
-    {| <h6 id="heading-6with-emphasis">Heading 6<em>with emphasis!</em></h6> |}]
+    {| <h6>Heading 6<em>with emphasis!</em></h6> |}]
 
 let%expect_test "construct lists" =
   show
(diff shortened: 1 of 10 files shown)


tests/expect_tests.ml
```diff
@@ -34,10 +34,10 @@ let%expect_test "construct inline elements" =
 
 let%expect_test "construct headings" =
   show Omd.Ctor.[ h 1 ~attrs:[ ("class", "my-class") ] [ txt "Heading 1" ] ];
-  [%expect {| <h1 id="heading-1" class="my-class">Heading 1</h1> |}];
+  [%expect {| <h1 class="my-class">Heading 1</h1> |}];
   show Omd.Ctor.[ h 6 [ txt "Heading 6"; em "with emphasis!" ] ];
   [%expect
-    {| <h6 id="heading-6with-emphasis">Heading 6<em>with emphasis!</em></h6> |}]
+    {| <h6>Heading 6<em>with emphasis!</em></h6> |}]
```
tatchi, line 40: I think that's a side effect of the new implementation. Since we now add the auto-ids at the time we parse the document, there's no way to have them auto-included here.

PR #307, opened 2024-03-12
Add canonical tags to a few types so odoc knows where to find them

jonludlam:
Currently the docs are either missing tags where they contain hidden items or linking to hidden modules. See ocaml/odoc#1091

### DIFF
--- a/src/ast_block.ml
+++ b/src/ast_block.ml
@@ -11,16 +11,21 @@ module InlineContent = struct
 end
 
 module List_types = struct
+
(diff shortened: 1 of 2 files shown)


jonludlam:
Before:
Screenshot 2024-03-12 at 14 41 53
After:
Screenshot 2024-03-12 at 14 37 11

shonfeder:
Much better!

shonfeder (approved):

